PR #153, opened 2018-05-18
We should be loading default config using workstation config loader

tyler-ball:
This must have gotten deleted at some point. Adding it back ensures we load any local `~/.chef/knife.rb` or `/etc/chef/client.rb` on user's workstation


tyler-ball:
@jonsmorrow @marcparadise Okay, I updated the code to reset the ChefConfig after we use the workstation config loader. This should prevent it from interfering with Policyfile creation later. What do you think about this now?

Issue #180, opened 2018-05-24
Configuration values are not respected when defaults are used in options


marcparadise:
## Description
When a custom configuration value is defined in config.toml, and the CLI option sets a default value to the configuration value, the customized default value is ignored. 

## Chef Workstation Version
0.1.120

## Platform Version
n/a 

## Replication Case
1. Set config option `connection.winrm.ssl` to `true`
2. run chef-run -h
3. output will contain a default cookbook repo path other than the provide value. 

## Client Output
```
grep ssl ~/.chef-workstation/config.toml
ssl=true
```
`chef-run -h`
-- snip
```
    -s, --[no-]ssl                     Use SSL for WinRM. Current default: false
```

tyler-ball:
I'm not sure this is going to be possible.

The reason it does not work now is that we evaluate the mixlib-cli options at class loading time, but the config file is not loaded until the `setup_cli` method is called. This means the help text description is generated before we read the config file.

If we want to change that, we would need to load the config file before parsing any of the mixlib-cli options. I think this could be done? But we would have to parse the ARGV ourselves. Or maybe we can figure out a way to load the mixlib-cli options after the class has been initialized?

Either way we do it, we need to load the config file into the `ChefRun::Config` object before we parse the options. But one of the options specifies the `--config-path`, so we would need to manually read that from the ARGVs ourselves if we want to load the config before generating the options descriptions.

tyler-ball:
I ran into another issue that is kind of the inverse of this.

If a user sets `--no-ssl` on the command line, but the default value is `ssl=true` we never persist that overridden value (`false`) anywhere. Everything works today because we pass the `config` object (lower case `c`, refers to the values we pulled from the parsed command line options) to the classes we need it.

But if a user later references the global `ChefRun::Config.connection.winrm.ssl` it will continue to return the default value (`true`) instead of the overridden value (`false`).

I think we could fix this by doing `ChefRun::Config.merge!(config)` to persist the CLI options back into the config global. But then we are really maintaining two different copies of config - the `config` hash and the `ChefRun::Config` global.

I think I would like to have a conversation with you @marcparadise where we decide if we can fix all of these config related issues at once. None of them seem very big and I think they overlap a lot of the same space.

PR #181, opened 2018-05-24
[SHACK-213] [GH-178] Add --protocol flag

### LINKED ISSUE #178: no need to type winrm://
It would be nice if I didn't have to type `winrm://` in front of every server I want to connect to. 

Possible solutions

1. Add a `--winrm` parameter
2. Let me default to winrm in a config file (useful if I have more windows than linux hosts)
3. Auto detect the remote host os type if not specified

marcparadise:
### Description
Not all customers use ssh for the majority of their hosts.
Currently, the only alternative we give is to require
the 'winrm://' prefix for each target.  This change adds
the flag --protocol to apply to all targets, and adds the
connection.default_protocol option

NOTE: the command line flag will work with this PR, but the configuration option 
will not work before #180 is fixed and merged. 

### Issues Resolved

Fixes #178 

### Check List

- [x] All tests pass

### DIFF
--- a/components/chef-run/lib/chef-run/target_resolver.rb
+++ b/components/chef-run/lib/chef-run/target_resolver.rb
@@ -20,7 +20,6 @@ require "chef-run/error"
 
 module ChefRun
   class TargetResolver
-    # IDeally, we'd base this on the actual current screen height
     MAX_EXPANDED_TARGETS = 24
 
     def initialize(unparsed_target, conn_options)
@@ -29,6 +28,7 @@ module ChefRun
       @conn_options = conn_options.dup
       @default_password = @conn_options.delete(:password)
       @default_user = @conn_options.delete(:user)
+      @default_proto = @conn_options.delete(:protocol) || ChefRun::Config.connection.default_protocol
     end
 
     # Returns the list of targets as an array of strings, after expanding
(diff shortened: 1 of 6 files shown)


components/chef-run/lib/chef-run/target_resolver.rb
```diff
@@ -20,14 +20,14 @@
...
   class TargetResolver
-    # IDeally, we'd base this on the actual current screen height
     MAX_EXPANDED_TARGETS = 24
 
     def initialize(unparsed_target, conn_options)
       @unparsed_target = unparsed_target
       @split_targets = unparsed_target.split(",")
       @conn_options = conn_options.dup
       @default_user = @conn_options.delete(:user)
+      @default_proto = @conn_options.delete(:protocol) || ChefRun::Config.connection.default_protocol
```
tyler-ball: This `|| ChefRun::Config.connection.default_protocol` should not be necessary since we are defaulting the connection options in the config

marcparadise: Good catch -  I was thinking defaulted options didn't come back, but it's only those without any default that don't.

marcparadise:
@tyler-ball I was in the process of making that requested change when I remembered why it fell back to default. `TargetResolver` doesn't "know" that it will always receive that option - it's not a required argument.  I'll update to make it one since that's the only thing that class needs external configuration for in the first place.

marcparadise:
@tyler-ball updates made

PR #183, opened 2018-05-24
[SHACK-210] [GH #180] Perform startup tasks outside of main CLI handling

### LINKED ISSUE #180: Configuration values are not respected when defaults are used in options

marcparadise:
In order to ensure that all configuration is loaded before we populate CLI options via Mixlib::CLI, 
this change separtes o

This resolves #180, in which changed configuration values were not recognized by the CLI when they were referenced before loading configuration - resulting in default values being propagated through the system. 

Changes: 
- Move tasks related to setup or initialization into a new Startup class
- chef-run executable now invokes Startup.run, which hands off control to CLI after all startup tasks are complete.

### DIFF
--- a/components/chef-run/bin/chef-run
+++ b/components/chef-run/bin/chef-run
@@ -9,14 +9,15 @@
 # You may obtain a copy of the License at
 #
 #     http://www.apache.org/licenses/LICENSE-2.0
-#
+
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+require "chef-run/startup"
 
-require "chef-run/cli"
-
-ChefRun::CLI.new(ARGV.clone).run
+# Perform initialization tasks then hand off control to
+# CLI to run the command.
+ChefRun::Startup.new(ARGV).run
--- a/components/chef-run/lib/chef-run/config.rb
+++ b/components/chef-run/lib/chef-run/config.rb
@@ -56,12 +56,17 @@ module ChefRun
         File.join(WS_BASE_PATH, "installation_id")
       end
 
+      def base_log_directory
+        File.dirname(log.location)
+      end
+
+      # These paths are relative to the log output path, which is user-configurable.
       def error_output_path
-        File.join(File.dirname(log.location), "errors.txt")
+        File.join(base_log_directory, "errors.txt")
       end
 
       def stack_trace_path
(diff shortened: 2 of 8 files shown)


components/chef-run/bin/chef-run
```diff
@@ -9,14 +9,24 @@
 # You may obtain a copy of the License at
 #
 #     http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
+
+# Unless required by applicable law or agreed to in writingtgtg, software
```
marcparadise: oops

components/chef-run/lib/chef-run/config.rb
```diff
@@ -80,16 +80,6 @@ def exist?
         File.exist? location
       end
 
-      def create_directory_tree
-        FileUtils.mkdir_p(File.dirname(default_location))
-        FileUtils.mkdir_p(File.dirname(stack_trace_path))
-        FileUtils.mkdir_p(telemetry_path)
-      end
-
-      def create_default_config_file
```
marcparadise: Moved behavior to Startup,   because a Configuration shouldn't be altering your file system.

components/chef-run/lib/chef-run/startup.rb
```diff
@@ -0,0 +1,144 @@
...
+      setup_logging
+
+      # Begin upload of previous session telemetry. (If telemetry is not enabled,
+      # in config the uploader will clean up previous session(s) without sending)
+      start_telemeter_upload
+
+      # Launch the actual chef-run behavior
+      start_chef_run
+    rescue ConfigPathInvalid => e
+      init_terminal
```
tyler-ball: Since we run `init_terminal` _first_ in the `run` method do we need to run it again here? Ditto to the other rescue below.

I realize this just ensures that our output call below will be successful, but it still feels redundant. Mostly it seems to protect us from a future situation where we are not calling `init_terminal` first in the `run` method

marcparadise:
@tyler-ball updated per your suggestion/question: https://github.com/chef/chef-workstation/pull/183/commits/cb06680c92bb77d9c2aadf616e5180f40ce074b2

PR #192, opened 2018-06-04
[SHACK-225] Merge CLI configs into global config object

tyler-ball:
The largest part of this diff is moving the existing CLI options into
cli_options.rb since cli.rb was getting pretty large.

### Description

If a user supplied `--ssl` on the command line our program would work because we pass the CLI specified `config` hash down into the classes where it is needed. But if someone later tried to access `ChefRun::Config.connection.winrm.ssl` it would return the default value of `false` instead of the user supplied value of `true`.

### Issues Resolved

SHACK-225

### Check List

- [x] All tests pass


components/chef-run/i18n/en.yml
```diff
@@ -32,7 +32,7 @@ cli:
   description: |
     Chef Run is a tool to execute ad-hoc tasks using Chef.
   creating_config: "Creating config file in %1."
-  default_config_location: "Location of config file. Defaults to %1"
+  default_config_location: "Location of config file. Default: %1"
```
tyler-ball: I found this wording issue while looking at the help text - brought it in line with other descriptions

components/chef-run/lib/chef-run/cli.rb
```diff
@@ -188,16 +91,16 @@ def handle_run_error(e)
 
     def perform_run
       parse_options(@argv)
-      if @argv.empty? || config[:help]
+      if @argv.empty? || parsed_options[:help]
```
tyler-ball: I really hate that mixlib-cli exposes the parsed options in a `config` method, so I wrote a wrapper and called it `parsed_options` - definitely open to a better name if someone thinks of one

components/chef-run/lib/chef-run/cli_options.rb
```diff
@@ -0,0 +1,146 @@
...
+          end
+          path
+        end)
+
+      klass.option :ssl,
+        long: "--[no-]ssl",
+        description:  T.ssl.desc(ChefRun::Config.connection.winrm.ssl),
+        boolean: true,
+        default: ChefRun::Config.connection.winrm.ssl,
+        proc: Proc.new { |val| ChefRun::Config.connection.winrm.ssl(val) }
```
tyler-ball: Here you can see an example of the pattern for writing values back to the global config. See the comment at the top of this module and make sure the explanation there makes sense.

components/chef-run/spec/integration/fixtures/chef_help.out
```diff
@@ -52,8 +52,8 @@ FLAGS:
     -p, --protocol <PROTOCOL>          The protocol to use for connecting to targets.
                                        The default is 'ssh', and it can be changed in config.toml by
                                        setting 'connection.default_protocol' to a supported option.
-    -s, --[no-]ssl                     Use SSL for WinRM. Current default: false
-    -s, --[no-]ssl-verify              Verify peer certificate when using SSL for WinRM
+        --[no-]ssl                     Use SSL for WinRM. Current default: false
```
tyler-ball: Another issue I saw - we were trying to use `-s` for both ssl and ssl verify

components/chef-run/lib/chef-run/cli_options.rb
```diff
@@ -0,0 +1,146 @@
...
+#   not want all CLI params merged back into the global config object.
+# We know that the config is already loaded from the file (or program defaults)
+#   because the `Startup` class was invoked to start the program.
+module ChefRun
+  module CLIOptions
+
+    T = ChefRun::Text.cli
+    TS = ChefRun::Text.status
+
+    def self.included(klass)
```
marcparadise: nice

PR #194, opened 2018-06-05
[SHACK-187] Error handling refactor

marcparadise:

### Description
This is a WIP.  

### Check List

- [ ] New functionality includes tests
- [ ] All tests pass
- [ ] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>

### DIFF
--- a/docs/architecture/errors.md
+++ b/docs/architecture/errors.md
@@ -161,10 +161,6 @@ structure makes it difficult to find out when and where certain clsases of
 errors occur (anything that's more 'compile' related). In addition, there's a bit of a feeling
 of "what's going to happen to this?" whenever you raise an exception.
 
-
-
-Target state:
-
 To meet the requirements above (which are not fully met now),
 proposed changes are:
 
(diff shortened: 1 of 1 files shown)


docs/architecture/errors.md
```diff
@@ -0,0 +1,209 @@
...
+Unknowns
+* how to best handle the mapping of external error types to internal ux-focused errors?
+   - currently we handle some of this black-box style in the global error handling,
+     while other parts get handled by specific classes that throw the errors, and still
+     others do so within the exceptions they create when handling expected failrue modes.
+
+Things to keep:
+0. (Maybe?) global handling and rendering for consistency
+  * this is a maybe because the global handling is also what leads to having
+    to do nested handling.
```
marcparadise: Overall I like the convenience of just being able to raise an error and know it'll get handled the right way for the operator. But the global handling is complicated to folllow because of: 
- telemetry execution blocks
- multi-target runs

docs/architecture/errors.md
```diff
@@ -0,0 +1,209 @@
...
+      - used for errors when the cause and remediation are clear, for example
+        bad option arguments to the CLI.
+      - maps to ChefRun::Error#show_stack == false && show_log == false
+
+#### One-off Formatting
+
+The Startup class runs outside of the main error handling, and does its own formatting
+of the errors it expects to see.
+
+
```
tyler-ball: I think I would like to see
```
#### Remote Chef Client Errors

When `chef-client` fails on the remote node we use a custom error handler to write run
information into a json file on the remote node. We fetch that JSON file and parse it locally on
the workstation to determine the cause of the error. Currently we send back the error text and
use a regular-expression based parser to map remote errors into informative local errors. In
the future we would like to get rid of this regex matching and instead send better information
in the formatted JSON.
```
or something like that

marcparadise: I missed this feedback when you posted it - thanks @tball 

I'll update with your suggestions.  The intent here was just to organize thoughts and make a plan to move forward -  not create a living reference to the arch.  Maybe I'll add something to the summary explaining that and a 'true as of mm/dd/yyyy' disclaimer.

PR #196, opened 2018-06-08
[SHACK-233] Do not overwrite existing Policyfile

tyler-ball:
### Description

When converging a recipe in an existing cookbook we don't want to blow
away a possible existing Policyfile because it could have custom sources
configured.

### Issues Resolved

SHACK-233

### Check List

- [x] All tests pass

### DIFF
--- a/components/chef-run/lib/chef-run/cli.rb
+++ b/components/chef-run/lib/chef-run/cli.rb
@@ -31,6 +31,7 @@ require "chef-run/telemeter"
 require "chef-run/telemeter/sender"
 require "chef-run/temp_cookbook"
 require "chef-run/text"
+require "chef-run/ui/terminal"
 require "chef-run/ui/error_printer"
 require "chef-run/version"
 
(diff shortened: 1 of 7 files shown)


components/chef-run/lib/chef-run/cli.rb
```diff
@@ -199,7 +200,11 @@ def perform_run
                                           config.delete(:protocol),
                                           config).targets
         temp_cookbook, initial_status_msg = generate_temp_cookbook(cli_arguments)
-        local_policy_path = create_local_policy(temp_cookbook)
+        local_policy_path = nil
+        UI::Terminal.render_job(TS.generate_policyfile.generating) do |reporter|
```
tyler-ball: Generating the Policyfile can sometimes take a few seconds, and right now it looks like the system hangs. So I also wanted to add a spinner around this process

components/chef-run/lib/chef-run/ui/error_printer.rb
```diff
@@ -236,6 +236,7 @@ def add_formatted_backtrace(out)
         if backtrace_length > cause_trace.length
           out.print "\t... #{backtrace_length - cause_trace.length} more"
         end
+        out.print "\n"
```
tyler-ball: Found an issue where stack traces we print were squashed together in instances where there are nested errors

PR #201, opened 2018-06-18
Our verify scripts should fail the pipeline if any of our commands fail

tyler-ball:
### Description

Our pipeline is passing `chef-run version` when it should not be.

### Issues Resolved

N/A

### Check List

- [x] All tests pass

Signed-off-by: tyler-ball <tyleraball@gmail.com>

### DIFF
--- a/ci/verify.bat
+++ b/ci/verify.bat
@@ -5,7 +5,9 @@ REM ; package is working correctly after install.
 
 REM ; chef-run version ensures our bin ends up on path and the basic ruby env is
 REM ; working.
-call chef-run version
+call chef-run --version
+IF %ERRORLEVEL% NEQ 0 goto :error
 
 REM ; Ensure our ChefDK works
-chef env
+call chef env
+IF %ERRORLEVEL% NEQ 0 goto :error
--- a/ci/verify.sh
+++ b/ci/verify.sh
@@ -1,8 +1,9 @@
 # This script is used in the jenkins part of our pipeline to verify our package
 # is working correctly after install.
+set -e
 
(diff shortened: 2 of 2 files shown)


ci/verify.bat
```diff
@@ -5,7 +5,16 @@ REM ; package is working correctly after install.
...
 REM ; chef-run version ensures our bin ends up on path and the basic ruby env is
 REM ; working.
-call chef-run version
+chef-run --version || goto :error
 
 REM ; Ensure our ChefDK works
 chef env
+
+REM ; Test
+chef-run version || goto :error
```
tyler-ball: Running a build through with this intentional failure right now to make sure this pattern works on Windows

tyler-ball: The unix tests are passing an ad-hoc run well, but Windows is printing the following:
```
chef-run: 0.1.136
Regular expression run condition: Expression=[^(?!windows).*$], Label=[windows-2012r2]
Run condition [Regular expression match] preventing perform for step [Execute shell]
```

It correctly prints the version from `chef-run --version` on line 8 above. But this line should be failing and it is not. I tried running a similar batch file on my Windows workstation and it seemed to fail appropriately. @tduffield or @scotthain any ideas?

tduffield: I do not unfortunately. If I find some time I can do some research, but maybe someone more familiar with Windows might know?

scotthain: Try an iteration of `chef-run --version || echo "Exited with error %errorlevel%" && exit /b` and the like!

PR #208, opened 2018-06-22
Support Automate self signed certs (short term)

tyler-ball:
### Description

With an Automate server using a self signed certificate trying to send run data to the data collector fails with an SSL error. In the short term to fix this we are transfering any trusted certificates on the Workstation to the target nodes. Users will need to run `knife ssl fetch https://automate-server` on their workstation for this to fix the issue.

In the long term we need to do a better job of surfacing this error, as well as fetching these certificates to the workstation automatically (if the user specifies they would like that) 

### Issues Resolved

Jira SHACK-255

### Check List

- [x] All tests pass


components/chef-run/spec/unit/cli_options_spec.rb
```diff
@@ -61,7 +61,7 @@ def parse(argv)
     expect(ChefRun::Config.connection.winrm.ssl).to eq(false)
     expect(ChefRun::Config.connection.winrm.ssl_verify).to eq(true)
     expect(ChefRun::Config.connection.default_protocol).to eq("ssh")
-    expect(ChefRun::Config.chef.cookbook_repo_paths).to eq(ChefConfig::Config[:cookbook_path])
+    expect(ChefRun::Config.chef.cookbook_repo_paths).to eq(ChefRun::Config.chef.cookbook_repo_paths)
```
jonsmorrow: Looks like the assert and eq are the same?

PR #211, opened 2018-06-25
Adding expeditor buffers

tyler-ball:
### Description

Keeps multiple PR merges from causing cascading builds.

### Issues Resolved

SHACK-223

### Check List

- [x] All tests pass


robbkidd:
Can we document why we would like to prevent multiple builds and why we're okay with a single, expedited version bump including up to 5 probably unrelated changes?

.expeditor/config.yml
```diff
@@ -18,23 +18,31 @@ github:
   # Delete the PR branch after successfully merged into release branch.
   delete_branch_on_merge: true
 
-# These actions are taken, in order they are specified, anytime a Pull Request is merged.
+buffers:
+  - build:
+      type: delay
+      workload_type: pull_request
+      delay: 600
```
marcparadise: Should we consider something like a 1hr delay? In the typical case it's not that urgent that we get updated packages. Perhaps if there's a way to force it anyway for urgent builds? If not, I agree 10m is better.

jonsmorrow: My vote would be start with 10 or even 5 and increase if we need to.

PR #213, opened 2018-06-26
[SHACK-256] We should be pulling cookbooks from locally configured coookbook repos

tyler-ball:
### Description

This fixes a customer issue where they do not want to specify their own
Policyfile.rb for all their local cookbooks. Instead we will populate
the :chef_repo for the Policyfile we generate with any of their
configured repo paths. Those paths can be specified in Chef Workstation
config, to the chef-run CLI or via their knife.rb.

### Issues Resolved

SHACK-256

### Check List

- [ ] New functionality includes tests
- [ ] All tests pass
- [ ] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>


components/chef-run/spec/unit/temp_cookbook_spec.rb
```diff
@@ -121,6 +127,21 @@
...
+          ChefRun::Config.chef.cookbook_repo_paths = %w{one two}
+          f = tc.generate_policyfile("foo", "bar")
+          expect(File.read(f)).to eq <<-EOD.gsub(/^ {10}/, "")
+          name "foo_policy"
+          default_source :chef_repo, "one"
+          default_source :chef_repo, "two"
+          default_source :supermarket
+          run_list "foo::bar"
+          cookbook "foo", path: "."
+          EOD
```
robbkidd: Could I convince you to update this heredoc for readability? Recommendations:

* indent the contents of the heredoc one more step so that, during a skim of the spec, it jumps out to the reader that there is only one expectation and that the expectation is _not_ followed by a series of commands
* use a [squiggly heredoc](https://ruby-doc.org/core-2.3.0/doc/syntax/literals_rdoc.html#label-Here+Documents) to eat the whitespace instead of a chained `gsub`
* give the heredoc a name meaningful in the context of the spec, like `<<~EXPECTED_POLICY`

🙏 😄

robbkidd: I tinkered with this spec in a local branch. This is what I came up with:

```ruby
context "when there are configured cookbook_repo_paths" do
  it "generates a policyfile in the temp cookbook" do
    ChefRun::Config.chef.cookbook_repo_paths = %w{one two}
    generated_policyfile = File.read(tc.generate_policyfile("foo", "bar"))
    expect(generated_policyfile).to eq <<~EXPECTED_POLICYFILE
      name "foo_policy"
      default_source :chef_repo, "one"
      default_source :chef_repo, "two"
      default_source :supermarket
      run_list "foo::bar"
      cookbook "foo", path: "."
    EXPECTED_POLICYFILE
  end
end
```

However, this style change to the spec would make it unlike the rest of the specs in this group of tests, which is beyond the scope of this expedited change. I say ship it and some time soon I'll offer up a PR with a style update for folks to consider.

tyler-ball: I had no idea about the squiggly heredoc - thank you! I'll get it changed for this spec and we can fix it in all the others during another pass

PR #217, opened 2018-06-28
Fix 'unexpected error' on auth failure; ensure auth user always shows on connect

marcparadise:

### Description
This fixes minor two bugs introduced with the fix for SHACK-272:

1. if  no username given and no default in .ssh/config, the username is
not shown in the connection message. It now shows as the train default,
'root'
2. search-and-replace error that popped up on general ssh connection
failures - missed a variable name.

### Issues Resolved
n/a

### Check List

- [x] All tests pass


components/chef-run/lib/chef-run/target_host.rb
```diff
@@ -81,12 +81,13 @@ def connect!
...
-      raise ConnectionFailure.new(e.cause || e, opts)
+      raise ConnectionFailure.new(e.cause || e, config)
     end
 
-    # Returns the user being used to connect
+    # Returns the user being used to connect. Defaults to root if not specified or
+    # defaulted in .ssh/config (for ssh connections), as set up in '#apply_ssh_config'.
     def user
-      config[:user]
+      config[:user] || "root"
```
tyler-ball: Super minor - does Train export this anywhere. EG, could we do `config[:user] || Train::Connection.default_user` instead of us hardcoding "root"?

marcparadise: Good idea,  I think we can. We'll have to create a discardable 'ssh' transport instance to get that value if we don't have it already, but that's no big deal.

marcparadise: Updated, looks like default_options was provided at the class-level so we're good without having to instantiate.

PR #232, opened 2018-07-27
[SHACK-304] Subscribe to ChefDK merges in expeditor

tyler-ball:
### Description

This means that every time a ChefDK PR is merged we will get a new build
of Chef Workstation. This new Chef Workstation build pulls ChefDK from
master for the omnibus build so effectively means a we will get a new
Chef Workstation build for every minor version bump (PR merge) of
ChefDK.

This could potentially include a build of ChefDK that does not pass the
pipeline since the ChefDK pipeline build occurs asynchronously from the
expeditor merge action.

We also update the omnibus definition to only include ChefDK since that
pulls Chef Apply in as a gem dependency now.

### Issues Resolved

SHACK-304

### Check List

- [ ] PR title is a worthy inclusion in the CHANGELOG
- [ ] `www/site/content/docs/` has been updated with any relevant changes:


omnibus/config/projects/chef-workstation.rb
```diff
@@ -37,10 +37,6 @@
 build_version Omnibus::BuildVersion.semver
 build_iteration 1
 
-override :bundler,        version: "1.16.1"
```
tyler-ball: These were duplicated below so I just got rid of them

omnibus/config/software/chef-dk.rb
```diff
@@ -30,6 +30,20 @@
 dependency "zlib"
 dependency "libarchive"
 
+#
```
tyler-ball: I copied the latest version of this file from the ChefDK repo

.expeditor/config.yml
```diff
@@ -36,3 +38,23 @@ merge_actions:
         - "Expeditor: Skip Build"
         - "Expeditor: Skip All"
       only_if: built_in:bump_version
+
+# I copied the `ignore_labels` from the chef-dk expeditor config
+# since the labels apply to that workload (not the chef-workstation
+# workload like the `merge_actions` block)
+subscriptions:
```
tyler-ball: @chef/expeditor-maintainers Does this look like a valid subscription? Because Chef Workstation includes ChefDK I would like to build Chef Workstation on every ChefDK PR merge.

I understand that this means Chef Workstation may get a bad ChefDK because the omnibus build of ChefDK may fail. But we rebuild that whole process in the Chef Workstation pipeline, so if ChefDK fails then Chef Workstation should also fail.

tduffield: Since ChefDK does some post-merge activities, what you probably want to subscribe to is the completion event: https://expeditor.chef.io/docs/reference/workloads/#completed-workloads. So your subscription would look like this: 

```
- workload: chef/chef-dk:master_completed:pull_request_merged:chef/chef-dk:master:*
   ...
```

tyler-ball:
@tduffield Got the subscription updated

PR #235, opened 2018-08-17
Capitalization issue in Windows package (wix)

tyler-ball:
### Description

Fixes it

### Issues Resolved

N/A

### Check List

### DIFF
--- a/omnibus/resources/chef-workstation/msi/source.wxs.erb
+++ b/omnibus/resources/chef-workstation/msi/source.wxs.erb
@@ -56,6 +56,7 @@
     <CustomActionRef Id="WixBroadcastSettingChange" />
     <CustomActionRef Id="WixBroadcastEnvironmentChange" />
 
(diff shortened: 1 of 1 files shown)


stuartpreston:
Yes, these directories need to be public properties (uppercase) - so this is one way to fix it but it will generate an ICE99 warning at compile time.  

I checked in Omnibus and we seem to ignore any warnings being generated (would give a return code of 204) - see https://github.com/chef/omnibus/blob/7ac8251d5025dce1f93c3da85b52f55c6a9b94ec/lib/omnibus/packagers/msi.rb#L71 

So, whilst this is a valid fix I think a better fix (avoiding warnings) that should be applied here and in ChefDK would be something like:

```
  <SetDirectory Id="WINDOWSVOLUME" Value="[WindowsVolume]"/>
  <Directory Id="TARGETDIR" Name="SourceDir">
    <Directory Id="WINDOWSVOLUME">
  ...
```
(ref: http://kentie.net/article/wixtipstricks/)

tyler-ball:
@stuartpreston Updated!

PR #245, opened 2018-09-19
[SHACK-358] update docs and navigation for chef.sh

marcparadise:
### Description

This restructures the CLI reference documents to be under the
heading "Tools Reference" and adds placeholders for key chefdk tools.

Previously, the "CLI Reference" heading was under "Chef Workstation" and
contained only the chef-run CLI reference. This change puts Workstation
tool CLI reference docs under a single heading to make them easier to
find.

The reference doc for chef-run is complete, but the other components currently
contain links over the their docs.chef.io (or inspec.io/docs)
counterparts until the docs migration effort is complete. 
capture
capture

### Check List


www/site/content/docs/tools-reference/berks.md
```diff
@@ -0,0 +1,12 @@
++++
+title = "Tools Reference - berks"
```
cheeseplus: Seems overkill to have `Tools Reference` on each page when that will be clearly indicated by the Nav

marcparadise: I went back and forth on that - what I actually wanted was to get the full name in the <head> title, "Chef Workstation Tools Reference - TOOLNAME", while the menu item and the page heading just have the tool name. 
I'll push the update to be just the tool name, and can open another card to straighten out titles on the browser titlebar (which will be only  TOOLNAME with that change), vs what's shown on the menu and in the content content.

marcparadise: Updated, new screencaps inbound

kagarmoe:
I'd like to just call it "Reference", in keeping with our other docs sites.

marcparadise:
@kagarmoe thanks, I've updated and refreshed the screen captures.

jonsmorrow:
@kagarmoe I'd like to get your sign off before we merge this.

PR #246, opened 2018-09-27
Add taskkill action to windows install

marcparadise:
### Description

This adds a 'taskkill' of 'Chef Workstation App.exe'
after the installer initializes.  This will prevent installations
from failing because the tray app executable is in use and
can't be replaced.

Signed-off-by: Marc A. Paradise <marc.paradise@gmail.com

### Check List

- [x] You have locally validated the change (pending build) 
- [x] NA: ~`www/site/content/docs/` has been updated with any relevant changes~


tyler-ball:
@marcparadise I updated this to follow something similar to https://stackoverflow.com/questions/11439767/how-to-kill-a-process-from-wix - just tested it and it looks good! Correctly closed CWA when trying to do a new install

PR #249, opened 2018-09-28
[SHACK-354] windows install: capture install path to registry 

marcparadise:
### Description
This captures the install and bin directories for Chef Workstation under `HKLM\SOFTWARE\Chef\Chef Workstation` as `InstallDir` and `BinDir`.  The tray application now uses `BinDir` to find the path to `ohai` when gathering platform info for update checks. 

### Check List
- [x] commit to use temp branch of chef-workstation-app is reverted 
- [x] na: ~ `www/site/content/docs/` has been updated with any relevant changes:~


omnibus/config/software/chef-workstation-app.rb
```diff
@@ -20,7 +20,7 @@
 license_file "LICENSE"
 
 source git: "https://github.com/chef/chef-workstation-app"
-default_version "master"
+default_version "SHACK-354/omnitruck-update-checks"
```
marcparadise: this'll be reverted once the branch merges

jonsmorrow: Can we get a revert pr queued up.

marcparadise: I was just going to interactive rebase/omit this commit rather than having a temp change in the merged history

marcparadise: And done, the branch is ready to go after the other PR merges.

tyler-ball:
@marcparadise I originally thought about doing this when we wanted to display the installed version of Chef Workstation in the app. But instead I was able to pull it from the uninstall information - https://github.com/chef/chef-workstation-app/blob/134522c7e7722c8d669cb5cad062b37af2353966/src/chef_workstation.js#L17

If we do this, can we also store the current version into the registry? That way we could update that code to pull the version from there rather than from the uninstall information (something we cannot control).

Also, is there anything we need to do to ensure these entries are removed on uninstall?

marcparadise:
@tyler-ball to also thought about getting version this way, but it's actually a bit of a challenge getting that before we're installed to insert the right value. 

However , I did include InstallDir - which lets us get the manifest path instead of querying the ininstall key. 

I will verify the removal, but I think it will do the right thing, because it's declared as  part of a feature that's also getting uninstalled.

PR #252, opened 2018-10-10
Slim down the package and install sizes

tas50:
RPM/Deb compression dropped DK package size by about 22% and the additional cleanup steps reduced package / install size by about 30%.

We'll want to run this through jenkins to confirm things, but it worked great on Chef and DK.


omnibus/config/software/chef-dk.rb
```diff
@@ -79,23 +79,14 @@
   appbundle "test-kitchen", lockdir: project_dir, gem: "test-kitchen", without: %w{changelog debug docs provisioning}, env: env
   appbundle "inspec", lockdir: project_dir, gem: "inspec", without: %w{deploy tools maintenance integration}, env: env
 
-  %w{chef-apply chef-dk chef-vault ohai opscode-pushy-client cookstyle dco berkshelf}.each do |gem|
+  %w{chef-dk chef-apply chef-vault ohai opscode-pushy-client cookstyle dco berkshelf}.each do |gem|
```
jonsmorrow: Was there a specific reason for the order change here? I think we were trying to keep workstation stuff at the front for code organization.

omnibus/config/software/chef-dk.rb
```diff
@@ -79,23 +79,14 @@
...
+  # Clear git-checked-out gems (most of this cleanup has been moved into the chef-cleanup omnibus-software definition,
+  # but chef-client still needs git-checked-out gems)
+  block "Delete bundler git installs" do
     gemdir = shellout!("#{install_dir}/embedded/bin/gem environment gemdir", env: env).stdout.chomp
-    remove_directory "#{gemdir}/cache"
     remove_directory "#{gemdir}/bundler"
   end
-
-  # Clean up docs
-  delete "#{install_dir}/embedded/docs"
```
jonsmorrow: This all moved to chef-cleanup i assume?

tas50: Yes

omnibus/config/software/chef-dk.rb
tas50: It was just a sync from DK where it's this order

tas50:
This now builds in Jenkins with the updated omnibus-software definition pulled in.

jonsmorrow:
looks like it saves about 30MB on linux: 157.92 MB -> 125.97 MB. Didn't see change on mac.

PR #255, opened 2018-10-10
Subscribe to Workstation App pr merges

jonsmorrow:
### Description

This change attempts to subscribe to merge events
on chef/chef-workstation-app and use that to manage
our dependency on the project.

### Check List

### DIFF
--- a/.expeditor/update_chef-workstation-app_to_latest.sh
+++ b/.expeditor/update_chef-workstation-app_to_latest.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+############################################################################
+# What is this script?
+#
+# Chef Workstation uses Expeditor to manage the bundled version of the
+# Chef Workstation App. Currently we always want to include the latest version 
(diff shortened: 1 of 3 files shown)


.expeditor/update_chef-workstation-app_to_latest.sh
```diff
@@ -0,0 +1,32 @@
...
+# Chef Workstation uses Expeditor to manage the bundled version of the
+# Chef Workstation App. Currently we always want to include the latest version 
+# of Chef Workstation App inside Workstation, so this script takes that version
+# and uses sed to insert it into the omnibus project definition. Then it commits
+# that change and opens a pull request for review and merge.
+############################################################################
+
+set -evx
+
+version=$(get_github_file $SCM_SETTINGS_GITHUB_REPO master VERSION)
```
tyler-ball: Since this script is pretty hard-coded now, why don't we just use `get_github_file chef-workstation-app master VERSION` instead of having that be dynamic. Is more straightforward that way, but less flexible if we modify this in the future

PR #266, opened 2018-10-12
Fix reversed condition for CWA lib detection

marcparadise:

### Description
This corrects a reversed `postinst` conditional that would cause the package install to tell the user that they could run chef-workstation-app when there were still deps to install. 

### Check List


marcparadise:
Added do-not-merge, I want to do some more testing of this.

PR #275, opened 2018-10-15
Adds chef-run user guide under Workstation

marcparadise:


www/site/content/docs/chef-workstation/chef-run-users-guide.md
```diff
@@ -0,0 +1,159 @@
++++
+title = "chef-run User's Guide"
```
marcparadise: Proposed Title - the title in THe Big Doc was "examples"

susanev: I think we can drop _User's_

www/site/content/docs/chef-workstation/chef-run-users-guide.md
```diff
@@ -51,7 +51,7 @@ To target WinRM you must specify the `winrm` protocol as part of the connection
 chef-run 'winrm://my_user:c0mplexP@ssword#!@host:5986' directory /tmp/foo
 ```
 
-WinRM connections only support password authentication. It will not read default information from the SSH config so all connection information must be specified on the command line. The port only needs to be specified if the target machine has it set to a non-default port (5986).
+WinRM connections only support password authentication and does not read default information from the SSH configuration. When using WinRM, specify all connection information on the command line. Only specify the connection port if be specified if the target machine uses a non-default port (default:5986).
```
susanev: not sure what you wanted here, but i dont think the "if be" is correct

www/site/content/docs/chef-workstation/chef-run-users-guide.md
```diff
@@ -1,5 +1,5 @@
 +++
-title = "chef-run User's Guide"
+title = "chef-run Guide"
```
susanev: do we want chef-run in back ticks here?

marcparadise: I'm not usre how that will render, I'll take a look

marcparadise: It just renders the literal backticks - both in the nav bar, and the page heading, so not in this case.

www/site/content/docs/chef-workstation/chef-run-users-guide.md
```diff
@@ -36,7 +36,7 @@ One choice for specifying your `chef-run` command is:
 chef-run host1 directory /tmp/foo
 ```
 
-To use password authentication instead of an identity file, specify it as part of the connection information or using the command line flag:
+To use password authentication instead of an identity file, specify the identity file location as part of the connection information or using the command line flag:
```
marcparadise: Should be: `specify the password as part of the connection information or using the command line flag`

(Which highlights why the change was needed in the first place :D)

PR #284, opened 2018-10-16
Bump DK version to v3.4.18

jonsmorrow:

### Check List


.expeditor/config.yml
```diff
@@ -43,9 +43,10 @@ merge_actions:
       only_if: built_in:bump_version
 
 subscriptions:
-  - workload: artifact_published:stable:chefdk:3*
-    actions:
-      - bash:.expeditor/update_chefdk_to_stable.sh
+  # Commented out while we have ChefDK pinned until after 3.4 release goes out
```
jonsmorrow: I think this should stay in... It only triggers when we promote dk to stable and we'd want to pick up when dk goes 3.4.x... Also it only creates the pr so we have a choice to merge.

PR #294, opened 2018-10-17
Update omnibus to use ruby-cleanup

tas50:
chef-cleanup has been renamed to ruby-cleanup since that's what it
actually is. This updates the omnibus definition to use the new
definition name.


jonsmorrow:
thanks @tas50

PR #306, opened 2018-10-22
Fix two additional links to be relative

jonsmorrow:
### Description

All links should now be relative. I also made download links use omintruck to
download directly.

### Check List

- ~[ ] PR title is a worthy inclusion in the CHANGELOG~
- [X] You have locally validated the change
- [X] `www/site/content/docs/` has been updated with any relevant changes:

### DIFF
--- a/www/site/content/docs/chef-workstation/getting-started.md
+++ b/www/site/content/docs/chef-workstation/getting-started.md
@@ -12,7 +12,7 @@ Chef Workstation gives you everything you need to get started with Chef. Ad-hoc 
 
 ## Install Chef Workstation
 
-If you have not installed Chef Workstation, download and install it from the [chef.sh](https://www.chef.sh) website.
+If you have not installed Chef Workstation, <a href="#" data-omnitruck-download="chef-workstation">download</a> and [install]({{< ref "../chef-workstation/install.md#installing" >}}) it.
 
 ## Check versions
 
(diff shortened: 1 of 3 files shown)


www/site/content/docs/chef-workstation/getting-started.md
```diff
@@ -12,7 +12,7 @@ Chef Workstation gives you everything you need to get started with Chef. Ad-hoc
 
 ## Install Chef Workstation
 
-If you have not installed Chef Workstation, download and install it from the [chef.sh](https://www.chef.sh) website.
+If you have not installed Chef Workstation, download and install it from the [chef.sh](/) website.
```
tyler-ball: This takes you to https://www.chef.sh/ but then users still need to click `Downloads` which takes them to downloads.chef.io - why not just go there?

jonsmorrow: yeah i can update it. I was trying not to change the functionality.

jonsmorrow: it should take you to localhost so you've got some caching. The only reason I can think of that they wanted to go to sh is the the button make sure they get the latest.

jonsmorrow: Actually thinking about this it's the same number of steps because the down load button figures out version and platform for the user.

PR #312, opened 2018-10-30
Introduce Release Notes process; Announce releases on discourse.chef.io

schisamo:
This change introduces the release notes process we have been
solidifying for Automate 2:

- Release Notes are captured in a simple wiki page on the project repository: https://github.com/chef/chef-workstation/wiki/Stable-Release-Notes
- These captured release notes are then automatically published to a known publicly accessible location when the promotion to `stable` is executed

Please remember release notes are not changelogs! The audience is our end-users, not other engineers. If you need a quick primer on what goes into good release notes, take a look at these excellent articles:
https://medium.com/@DigitalGov/the-life-changing-magic-of-writing-release-notes-4c460970565
https://www.prodpad.com/blog/writing-release-notes/

The published release notes are available at the following URLs:
```
https://packages.chef.io/release-notes/stable/chef-workstation/latest.md
https://packages.chef.io/release-notes/chef-workstation/<VERSION>.md
```
Additionally, we'll also begin to automatically announce release on discourse.chef.io. Here is an example announcement for Automate 2 which is produced in the same way:
https://discourse.chef.io/t/automate-2-version-20181020020209-released/13901

### DIFF
--- a/.expeditor/announce-release.sh
+++ b/.expeditor/announce-release.sh
@@ -12,23 +12,9 @@ $(cat release-notes.md)
 ---
 ## Get the Build
 
-As always, you can download binaries directly from [downloads.chef.io](https://downloads.chef.io/$product_key/$version) or by using the \`mixlib-install\` :
+If you are running the experimental application you can download this version from the menu after the app next update check. You can also download binaries directly from [downloads.chef.io](https://downloads.chef.io/$product_key/$version).
 
-\`\`\`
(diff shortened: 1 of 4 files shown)


tas50:
Is there a process in place for a situation where we end up with an N-1 release or are we just hoping we don't have to do that?

jonsmorrow:
@tas50 on workstation I am really hoping we will not do n-1 like we do for dk.

PR #331, opened 2019-01-15
Pin delivery-cli version and add subscription

jonsmorrow:
### Description

Now that A2 includes workflow there may be features added to delivery-cli.
In the past we always just pulled master whenever we built workstation, but
this means we only pick up delivery-cli changes when we build for other reasons.
This change pins the version and adds an expeditor subcription that updates a
new version override in our omnibus project definition so we can stay up
to date.

### Issues Resolved

None.

### Check List

- [ ] `www/site/content/docs/` has been updated with any relevant changes:

### DIFF
--- a/.expeditor/update_delivery-cli_to_latest.sh
+++ b/.expeditor/update_delivery-cli_to_latest.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+############################################################################
+# What is this script?
+#
+# Chef Workstation uses Expeditor to manage the bundled version of the Delivery CLI.
+# Currently we always want to include the latest version of Delivery CLI inside
+# Workstation, so this script takes that version and uses sed to insert
+# it into the omnibus project definition. Then it commits that change
+# and opens a pull request for review and merge.
+############################################################################
+
+set -evx
+
+branch="expeditor/delivery_cli_${VERSION}"
+git checkout -b "$branch"
+
+sed -i -r "s/override :\"delivery-cli\",\s+version: \"v[^\"]+\"/override :\"delivery-cli\", version: \"v${VERSION}\"/" omnibus/config/projects/chef-workstation.rb
+
+git add .
+
+# give a friendly message for the commit and make sure it's noted for any future audit of our codebase that no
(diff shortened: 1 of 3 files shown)


marcparadise:
I think this will need a new doc page or link to add the delivery CLI to reference docs.

jonsmorrow:
I think we still don't want to highlight delivery-cli since it's optional and not suggested path in A2.

jonsmorrow:
http://manhattan.cd.chef.co/job/chef-workstation-trigger-ad_hoc/177/downstreambuildview/

.expeditor/update_delivery-cli_to_latest.sh
```diff
@@ -0,0 +1,30 @@
...
+# it into the omnibus project definition. Then it commits that change
+# and opens a pull request for review and merge.
+############################################################################
+
+set -evx
+
+branch="expeditor/delivery_cli_${VERSION}"
+git checkout -b "$branch"
+
+sed -i -r "s/override :\"delivery-cli\",\s+version: \"v[^\"]+\"/override :\"delivery-cli\", version: \"v${VERSION}\"/" omnibus/config/projects/chef-workstation.rb
```
tyler-ball: Looks like there is currently no `v` on the version string - `"0.0.45"` not `"v0.0.45"`

marcparadise: Touching chef-workstation.rb will force a new full/uncached build; delivery-cli is pretty high up the stack, would it make sense to move this override (any maybe others getting auto-updated) to an external file? An example: 

https://github.com/chef/chef-server/blob/master/omnibus/config/projects/chef-server.rb#L33
and 
https://github.com/chef/chef-server/blob/master/omnibus_overrides.rb

tyler-ball: `sed -i -r "s/override\s+\"delivery-cli\",\s+version:\s+\"[^\"]+\"/override \"delivery-cli\", version: \"${VERSION}\"/" omnibus/config/projects/chef-workstation.rb`

jonsmorrow: Fixed in lates commit.

PR #332, opened 2019-01-16
New the new expeditor labels

tas50:
We migrated to these new labels already. This makes sure that expeditor still does the right thing.


tyler-ball:
And I get to use a label when merging this! Thanks @tas50 !

PR #353, opened 2019-02-27
Update omnibus overrides to match ChefDK.

jonsmorrow:

### DIFF
--- a/omnibus/config/software/chef-dk-powershell-scripts.rb
+++ b/omnibus/config/software/chef-dk-powershell-scripts.rb
@@ -25,7 +25,8 @@ license :project_license
 build do
   block "Install windows powershell scripts" do
     # Copy the chef gem's distro stuff over
-    chef_gem_path = File.expand_path("../..", shellout!("#{install_dir}/embedded/bin/gem which chef").stdout.chomp)
+    #   env: with_embedded_path required for rubygems > 2.7.6
+    chef_gem_path = File.expand_path("../..", shellout!("#{install_dir}/embedded/bin/gem which chef", env: with_embedded_path).stdout.chomp)
 
     chef_module_dir = "#{install_dir}/modules/chef"
     create_directory(chef_module_dir)
--- a/omnibus/Gemfile
+++ b/omnibus/Gemfile
@@ -1,11 +1,11 @@
 source "https://rubygems.org"
 
 # Install omnibus
-gem "omnibus", git: "https://github.com/chef/omnibus.git"
+gem "omnibus", git: "https://github.com/chef/omnibus.git", branch: "master"
 
 # Use Chef's software definitions. It is recommended that you write your own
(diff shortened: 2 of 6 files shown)


omnibus/config/software/chef-dk-powershell-scripts.rb
```diff
@@ -25,7 +25,8 @@
 build do
   block "Install windows powershell scripts" do
     # Copy the chef gem's distro stuff over
-    chef_gem_path = File.expand_path("../..", shellout!("#{install_dir}/embedded/bin/gem which chef").stdout.chomp)
+    #   env: with_embedded_path required for rubygems > 2.7.6
+    chef_gem_path = File.expand_path("../..", shellout!("#{install_dir}/embedded/bin/gem which chef", env: with_embedded_path).stdout.chomp)
```
jonsmorrow: @tas50 Would like your thoughts on this.

jonsmorrow:
@tas50 Please take another look at this now that we got it building.

omnibus/config/software/chef-dk-powershell-scripts.rb
tas50: Do we do this on the chef side?

jonsmorrow: not in dk. but it's the only thing we could find to have the lookup work... My guess is it has something to do with dk having a gemspec direct dependency on chef, but for us it's transitive.

jonsmorrow: @tas50 ^

lamont-granquist: the with_embedded_path does raise a bit of an alarm bell.

the shellout helper in omnibus should clean the env vars up and break out of the outer bundle, etc.   so this should be equivalent to `/opt/chefdk/embedded/bin/gem which chef` done cleanly, which should work.

that smells a bit like there's a new env var that isn't getting cleaned up by the omnibus helpers in 2.7.6, or that something happened to break the binstubs.

lamont-granquist: @jonsmorrow we don't have to do this on the chef/chef side because we're building that gem out of `/var/cache/omnibus/src/chef`, while here you've just `gem install`'d so you only have the chef sources in the ruby library.

and this is cleaner than what has been done in the past.  omnibus should most likely have a helper to make this less awful.

still i'm worried the need for `with_embedded_path` in this particular diff might be indicative of a larger bug.

lamont-granquist: i think we do something to fix up the shebangs in the binstubs in embedded, and that might be going wrong with new bundler/rubygems?

marcparadise:
This change isn't needed, builds are passing with latest omnibus.

PR #376, opened 2019-05-28
Add a chefworkstation docker container

tas50:
We have one for dk. Even though on the CLI it's the same we should include that.

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -8,19 +8,18 @@ product_key:
 slack:
   notify_channel: chef-ws-notify
 
+docker_images:
+  - chefworkstation
+
 github:
-  # The file where the MAJOR.MINOR.PATCH version is kept. The version in this file
-  # is bumped automatically via the `built_in:bump_version` merge_action.
-  version_file: "VERSION"
-  # The file where our CHANGELOG is kept. This file is updated automatically with
-  # details from the Pull Request via the `built_in:update_changelog` merge_action.
-  changelog_file: "CHANGELOG.md"
-  # Delete the PR branch after successfully merged into release branch.
+  # This deletes the GitHub PR branch after successfully merged into the release branch
   delete_branch_on_merge: true
   # allow bumping the minor release via label
   minor_bump_labels:
     - "Expeditor: Bump Version Minor"
-
+  # allow bumping the major release via label
+  major_bump_labels:
+    - "Expeditor: Bump Version Major"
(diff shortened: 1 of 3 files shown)


tas50:
Yeah this is really just for marketing sake. We can move everything off the DK containers

Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM ubuntu:18.04
+MAINTAINER Chef Software, Inc. <docker@chef.io>
+
+ARG CHANNEL=stable
+ARG VERSION=0.2.53
+ENV DEBIAN_FRONTEND=noninteractive \
+    PATH=/opt/chefworkstation/bin:/opt/chefworkstation/embedded/bin:/root/.chefdk/gem/ruby/2.6.0/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```
tyler-ball: We install to `/opt/chef-workstation`. I'll update this and rebase on master.

.expeditor/config.yml
```diff
@@ -8,19 +8,18 @@ product_key:
 slack:
   notify_channel: chef-ws-notify
 
+docker_images:
+  - chefworkstation
```
tyler-ball: What is the reason for this over `chef-workstation`?

tas50: it could go either way. Now is our chance

tyler-ball: I vote `chef-workstation` since it is easier to read

tas50: Push away

tas50:
@tduffield what am I missing in the expeditor config here?

.expeditor/config.yml
```diff
@@ -49,13 +49,21 @@ pipelines:
...
+    actions:
+      - built_in:build_docker_image
+  - workload: artifact_published:current:chef-workstation:{{version_constraint}}
+    actions:
+      - built_in:tag_docker_image
   - workload: artifact_published:stable:chef-workstation:{{version_constraint}}
     actions:
       - built_in:rollover_changelog
+      - bash:.expeditor/update_dockerfile.sh
+      - built_in:tag_docker_image
```
tduffield: Your actions here are out of order. `tag_docker_image` is a post-commit action, while bash actions are (by default) pre-commit actions. To address the expeditor config issue you'll want to move this action to the end of the list (or add a `post_commit: true`) to the subsequent bash actions.

PR #379, opened 2019-05-30
Copying in the latest ChefDK 3 omnibus definitions to fix the Workstation build

tyler-ball:
### Description

Getting ready for Workstation 0.3 release

### Issues Resolved

N/A

### Check List


omnibus/config/projects/chef-workstation.rb
```diff
@@ -67,17 +67,18 @@
 dependency "rust-uninstall"
 
 dependency "chef-dk"
+dependency "chef-dk-gem-versions"
 
 dependency "gem-permissions"
 dependency "rubygems-customization"
+dependency "shebang-cleanup"
```
tyler-ball: I'm so sad it took me a week to figure this out. Any idea why we did not include this dependency originally @jonsmorrow @marcparadise @robbkidd ?

omnibus/config/projects/chef-workstation.rb
```diff
@@ -67,17 +67,18 @@
...
 
 if windows?
   dependency "chef-dk-env-customization"
   dependency "chef-dk-powershell-scripts"
 end
 
 dependency "version-manifest"
-dependency "clean-static-libs"
```
tas50: ?

tyler-ball: Copied from https://github.com/chef/chef-dk/blob/v3.11.1/omnibus/config/projects/chefdk.rb#L69

tas50: Seems like we want that but if we're not doing it in DK we should nuke it for now

tyler-ball: I searched for a while through a few different means and could not figure out when that line got removed. I think it got dropped during crazytown days.

tas50: https://github.com/chef/omnibus-software/pull/1058 this cleans it up by adding the old logic into our new ruby cleanup def

PR #400, opened 2019-07-09
Build Chef Workstation as a self contained project w/o depending on the ChefDK project

marcparadise:
This change removes our dependency on ChefDK, and moves ownership of including the ChefDK tools into Chef Workstation itself.

* ChefDK dependency removed
  * Replaced with ChefCLI for the `chef` portion 
  * Workstation now owns the gem packaging and builds for Workstation.
* New `gems` component added to chef-workstation, which contains the One True Gemfile/.lock that defines all gems we distribute with chef-workstation. If it is not defined in components/gems/Gemfile, it will not be shipped.
  * `rake update` task for updating the lock
* New `gems` software definition that builds the `gems` component above
* ChefDK patch files have been removed

I'll take this out of Draft status once omnibus builds have been verified. 

Remaining work (separate)
- rename chef-dk software components appropriately
- sync up ChefCLI with patches
  - we will need to make sure no changes are needed in ChefCLI, but what we have now is pretty close to what were were patching into DK originally in the case of CLI output.
- make sure version updates are handled correctly
- update gem versions where the updates are to different versions than DK 
- talk about `chef` command version output, `verify` command output, `env` behaviors
- remove components we don't intend to ship (dco, spork, cucumber)

## Types of changes
- [x] Chore (non-breaking change that does not add functionality or fix an issue)


components/gems/Gemfile
```diff
@@ -0,0 +1,146 @@
...
+  # float we'll end up with 2 copies shipped in Workstation.
+  # When we bump Ruby we need to look at these pins and adjust them
+  gem "rake", "<= 12.3.2"
+  gem "rdoc", "<= 6.0.1"
+  gem "minitest", "<= 5.10.3"
+
+  gem "pry"
+  gem "yard"
+  gem "guard"
+  gem "cookstyle", "~> 4.0"
```
tas50: This should be 5.0. Did this file come from the latest DK?

components/gems/Gemfile
```diff
@@ -0,0 +1,146 @@
...
+# We equality pin the chef gem itself to assert which version we're shipping.
+group(:omnibus_package) do
+  gem "appbundler"
+  gem "berkshelf", ">= 7.0.5"
+  gem "chef-vault", ">= 3.4.1"
+  # Expeditor manages the version of chef released to Rubygems. We only release 'stable' chef
+  # gems to Rubygems now, so letting this float on latest should always give us the latest
+  # stable release. May have to re-pin around major version bumping time, or during patch
+  # fixes.
+  gem "chef", "= 15.0.300"
```
tas50: This should be 15.1

components/gems/Gemfile
```diff
@@ -0,0 +1,146 @@
...
+
+  # All of the following used to be software definitions we included:
+  gem "knife-spork"
+  gem "mixlib-install"
+  gem "nokogiri"
+  gem "pry-byebug"
+  gem "pry-remote"
+  gem "pry-stack_explorer"
+  gem "rb-readline"
+  gem "rubocop"
```
tas50: this can go away

components/gems/Gemfile
```diff
@@ -0,0 +1,146 @@
...
+  # For Delivery build node
+  gem "chef-sugar"
+  gem "mixlib-versioning"
+  gem "artifactory"
+  gem "opscode-pushy-client", ">= 2.99"
+  gem "ffi-rzmq-core"
+  gem "knife-push"
+
+  # All of the following used to be software definitions we included:
+  gem "knife-spork"
```
tas50: We should probably take this opportunity to remove this

components/gems/Gemfile
```diff
@@ -0,0 +1,146 @@
...
+  gem "knife-tidy", ">= 1.2.0"
+  gem "knife-windows", ">= 1.9.1"
+  gem "knife-opc", ">= 0.4.0"
+  gem "knife-vsphere", ">= 2.1.1"
+  gem "mixlib-archive", ">= 0.4.16"
+  gem "ohai", ">= 14"
+  gem "net-ssh", ">= 4.2.0"
+  gem "test-kitchen", ">= 1.23"
+  gem "listen"
+  gem "dco"
```
tas50: We should probably take this opportunity to remove this

components/gems/Gemfile
```diff
@@ -0,0 +1,146 @@
...
+  gem "knife-ec2", ">= 0.19.10"
+  gem "knife-google", ">= 3.3.3"
+  gem "knife-tidy", ">= 1.2.0"
+  gem "knife-windows", ">= 1.9.1"
+  gem "knife-opc", ">= 0.4.0"
+  gem "knife-vsphere", ">= 2.1.1"
+  gem "mixlib-archive", ">= 0.4.16"
+  gem "ohai", ">= 14"
+  gem "net-ssh", ">= 4.2.0"
+  gem "test-kitchen", ">= 1.23"
```
tas50: We should change this to >= 2

components/gems/Gemfile
```diff
@@ -0,0 +1,146 @@
...
+  gem "kitchen-dokken", ">= 2.6.7"
+  gem "kitchen-google", ">= 2.0.0"
+  gem "kitchen-hyperv", ">= 0.5.1"
+  gem "kitchen-inspec", ">= 1.0"
+  gem "kitchen-vagrant", ">= 1.4"
+  gem "knife-acl", ">= 1.0.3"
+  gem "knife-ec2", ">= 0.19.10"
+  gem "knife-google", ">= 3.3.3"
+  gem "knife-tidy", ">= 1.2.0"
+  gem "knife-windows", ">= 1.9.1"
```
tas50: This should be >= 3

components/gems/Gemfile
```diff
@@ -0,0 +1,146 @@
...
+  gem "kitchen-azurerm", ">= 0.14"
+  gem "kitchen-ec2", ">= 3.0", "< 4"
+  gem "kitchen-digitalocean", ">= 0.10.0"
+  gem "kitchen-dokken", ">= 2.6.7"
+  gem "kitchen-google", ">= 2.0.0"
+  gem "kitchen-hyperv", ">= 0.5.1"
+  gem "kitchen-inspec", ">= 1.0"
+  gem "kitchen-vagrant", ">= 1.4"
+  gem "knife-acl", ">= 1.0.3"
+  gem "knife-ec2", ">= 0.19.10"
```
tas50: This should be >= 1.0

components/gems/Gemfile
```diff
@@ -0,0 +1,146 @@
...
+  gem "mixlib-install"
+  gem "nokogiri"
+  gem "pry-byebug"
+  gem "pry-remote"
+  gem "pry-stack_explorer"
+  gem "rb-readline"
+  gem "rubocop"
+  gem "winrm-fs"
+  gem "winrm-elevated"
+  gem "cucumber"
```
tas50: I think we're safe to remove cucumber at this point as long as we call it out

components/gems/Gemfile
marcparadise: I was going to hold off on removals and make them separate changes; I wanted to start with "here's a build that's just like the build we had before" and ensure everything is behaving.

marcparadise:
I'll re-sync the Gemfile out of chef-dk - my last sync-up was probably wed-thu of last week.

components/gems/Rakefile
```diff
@@ -0,0 +1,31 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+task :default do
+  puts "There is no default task. You probably want 'bundle exec rake update'"
+end
+
+desc "Update the One True Gemfile.lock in this directory with the latest dependencies."
+task :update do
```
tas50: We probably want the other update tasks so we can do both the gemfile.lock and the omnibus.lock in one task

components/gems/README.md
```diff
@@ -0,0 +1,11 @@
+The `gems` component is used by appbundler
+to build all of the gems we ship as part of
+Chef Workstation.
+
+Gemfile.lock in this directory is the One True Source
+of shipped gems.
+
+To update the Gemfile.lock, run the rake task:
+
+`bundle exec rake dependencies:update_gemfile_lock`
```
marcparadise: s/b `bundle exec rake update`

components/gems/Rakefile
marcparadise: I don't think we do in this case - it worked at the top-level of chef-dk because you were updating all deps of chef-dk. 

It doesn't fit as well here, because the dep update is occurring within a single component that's not connected to the build-time environment''s dependencies for the building package.

omnibus/config/projects/chef-workstation.rb
```diff
@@ -78,11 +83,10 @@
   dependency "chef-dk-powershell-scripts"
 end
 
+
 dependency "version-manifest"
 dependency "openssl-customization"
 
-dependency "stunnel" if fips_mode?
```
tyler-ball: Hmm - seems like not supporting a fips distribution of Chef Workstation is going to break some of our customers. I realize we don't have the pipeline right now but this seems like an issue we should address.

marcparadise: The path we've followed in other products is to make every build fips-capable, and control the behavior via feature flag.  If we determine we need FIPs support in DK, I'm hoping we're able to do it the same way.

marcparadise: Got clarification from @robbkidd - this will be true for any fips-supported platform (rhel, windows) and not based on build configuration. I'll keep it in.

PR #401, opened 2019-07-09
Fix the chef --version command showing the old DK version

tas50:
This all goes away when we switch to the Chef CLI. Then we'll need to find another way to do it all.


tyler-ball:
Where does the logic that swaps in `$CHEF_WS_VERSION$` live? I haven't seen that before

tas50:
I think that's in the workstation omnibus bit. This is straight from the contents of the old patch so we have logic somewhere utilizing it.

PR #410, opened 2019-07-15
Wire up Chefstyle in Buildkite

tas50:
Wire up chefstyle testing and add a rakefile for the style task.

### DIFF
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -1,14 +1,23 @@
 steps:
-  - label: ":hugo:"
-    command:
-      - .expeditor/buildkite/hugo_lint.sh
-      - cd www && make lint
-    timeout_in_minutes: 10
-    retry:
-      automatic:
-        limit: 1
-    expeditor:
-      accounts:
-        - github
-      executor:
-        docker:
+- label: chefstyle-tests-ruby-2.6
+  command:
+    - export USER="root"
+    - bundle install --jobs=7 --retry=3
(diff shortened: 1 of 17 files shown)


Rakefile
```diff
@@ -0,0 +1,11 @@
+begin
+  require "chefstyle"
+  require "rubocop/rake_task"
+  desc "Run Chefstyle tests"
+  RuboCop::RakeTask.new(:style) do |task|
+    task.options += ["--display-cop-names", "--no-color"]
+  end
+rescue LoadError
+  puts "chefstyle gem is not installed. bundle install first to make sure all dependencies are installed."
```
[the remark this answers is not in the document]
tas50: I think traditionally we wanted to be able to exclude the stuff, but that probably doesn't make sense here. It all goes back to github changelog generator where we wanted to be able to skip that group but still run style w/o issues. It can probably go away at this point or yeah reraise.

.expeditor/verify.pipeline.yml
```diff
@@ -1,14 +1,23 @@
...
+  command:
+    - .expeditor/buildkite/hugo_lint.sh
+    - cd www && make lint
+  timeout_in_minutes: 10
+  retry:
+    automatic:
+      limit: 1
+  expeditor:
+    accounts:
+      - github
```
tduffield: Why do we need GitHub credentials for an open source project?

tas50: 🤷‍♂ chef.sh stuff

PR #417, opened 2019-07-18
Remove cucumber, dco, knife-spork

marcparadise:
These tools are not used by Workstation, and do not fit into any recommended workflow. 


components/gems/Gemfile
```diff
@@ -81,7 +81,6 @@ group(:omnibus_package) do
   gem "net-ssh", ">= 4.2.0"
   gem "test-kitchen", ">= 2.0"
   gem "listen"
-  gem "dco"   # TODO - dksplit - Planned for removal
```
robbkidd: I know that many (most? all?) of the Workstation project maintainers probably have git configured to do the commit sign-offs. I'm not so sure about the rest of the community. Did I miss some conversations about its utility?

marcparadise: The dco gem was included undocumented to ease the transition to the DCO process; the initial blog post details the instructions for configuring git and signing commits, but does not refer anyone to the dco tool to manage this. 

Given that the transition to DCO has gone smoothly, and the tool itself is not publicized in DK, there seems little value in continuing to maintain it in Workstation.

robbkidd: Fair enough.

marcparadise:
This bump to chef-cli should see verification passing.

PR #420, opened 2019-07-25
WIP - moving verification tests into chef-workstation

marcparadise:


omnibus/omnibus-test.ps1
```diff
@@ -1,40 +0,0 @@
-# Stop script execution when a non-terminating error occurs
```
marcparadise: This'll get replaced with verify-install.ps1 that is equivalent to verify-install.sh

omnibus/tests/verify-install.sh
```diff
@@ -0,0 +1,363 @@
...
+echo "foodcritic"
+foodcritic -V
+echo "inspec"
+inspec version
+
+
+echo "Verifying berkshelf"
+export_gem_location_for "berkshelf"
+pushd "$gem_location"
+  bundle install --quiet --with=development
```
lamont-granquist: i don't think you need the `--with` because of how appbundler renders the Gemfile.  i don't think you get any choices here about with or without at all.

omnibus/tests/verify-install.sh
```diff
@@ -0,0 +1,363 @@
...
+  echo "Running smoke test for test-kitchen"
+  export_temp_dir
+  pushd "$temp_dir"
+  #   # NOte that now we're directly running our packaged kitchen, and not via the bundle isntall above.
+  kitchen init --create-gemfile
+  popd
+  cleanup_temp_dir
+popd
+
+echo "Verifying policifile provisioning"
```
lamont-granquist: *policyfile 😄

marcparadise:
Replacement PR coming soon, taking an approach that doesn't involve maintaining two sets of identical tests (win/linux) separately.

PR #424, opened 2019-08-06
Update InSpec, kitchen-vagrant, and kitchen-azurerm to the latest

tas50:
Also require a modern berkshelf


tas50:
@afiune fixed

PR #454, opened 2019-08-26
Build on 10.13 but test on 12, 13, and 14

tas50:
We support macOS 10.12 still, but electron-builder no longer notorizes
on 10.12. We need to build on 10.13 and test on all versions.


tas50:
Closing out since @jaymalasinha got this in #455

PR #477, opened 2019-09-06
Fix unzip of chef-workstation-app (Electron-App)

### LINKED ISSUE #474: Post install we should delete the zipped copy of the workstation app
We currently leave the zipped up copy of the workstation app behind in the /opt/chef-workstation directory. This makes that install directory 63 megs bigger than it needs to be. We should delete that.
Screen Shot 2019-09-05 at 1 58 07 PM

afiune:
Electron changed the folder it zips the app into, needs to change unzip correspondingly

Also closes #474 


tas50:
Can we get a test added to workstation so we know we're actually installing the app? This would have been a really nasty regression if we had missed it.

afiune:
We are still trying to figure out the problem, please hold on merge. 😉

tyler-ball:
Lets potentially get https://github.com/chef/chef-workstation/issues/474 fixed as part of this PR

omnibus/package-scripts/chef-workstation/postinst
```diff
@@ -44,11 +44,11 @@ if is_darwin; then
   # within the larger Chef Workstatoin package, and will not include it in the Chef Workstation
   # package directly.
   echo "Moving Chef Workstation App to the Applications folder"
-  pushd $INSTALLER_DIR/components/chef-workstation-app
+  pushd "$INSTALLER_DIR/components/chef-workstation-app"
   unzip chef-workstation-app-mac.zip
   sudo rm -rf "/Applications/Chef Workstation App.app"
-  sudo mv "mac/Chef Workstation App.app" /Applications
-  rm -r mac
```
afiune: I think that this `rm` command made the `postinst` script to stop processing the rest of the commands in the script. Maybe this is the solution 🤞

afiune:
Waiting on this build to finish and then test the installer: https://buildkite.com/chef/chef-chef-workstation-master-omnibus-adhoc/builds/63

I have already added a test for MacOS to verify that we are installing the **'Chef Workstation App'**, but it is missing something similar for Windows.

afiune:
You got this Adhoc Build!!!
tenor-50737593

afiune:
I have verified that we have fixed the major bug introduced with the upgrade of Electron, now the Chef Workstation App continues to be installed on the system!  

tenor-73552396


Merging!

PR #481, opened 2019-09-11
Update gem-version-manifest software definition

afiune:
Let us try to remove all ChefDK wording from the Chef Workstation repo.


afiune:
Open Question: Do we really need to generate some of these scripts at build time? 🤔

PR #493, opened 2019-09-14
Update git for non-Windows platforms to 2.23.0

tas50:
This resolves CVE-2017-14867


tas50:
Confirmed this builds via adhoc

PR #505, opened 2019-09-16
Update knife-vsphere and kitchen-vcenter to the latest

tas50:


.expeditor/config.yml
```diff
@@ -176,7 +176,7 @@ subscriptions:
   - workload: ruby_gem_published:knife-vsphere-*
     actions:
        - bash:.expeditor/update_dep.sh
-  - workload: ruby_gem_published:knife-vsphere-*
+  - workload: ruby_gem_published:knife-vcenter-*
```
tas50: vsphere was in there twice and vsphere wasn't in there

PR #508, opened 2019-09-17
Update nodejs to 10.16.3

tas50:
A giant pile of bug fixes


tas50:
It built, but I need to grab that build and figure out if it actually works locally.

tyler-ball:
@tas50 We verified this works on Ubuntu, Windows and MacOS!

PR #509, opened 2019-09-17
Simplify the builder cookbook

tas50:
I'm not sure if this is used anymore but it certainly doesn't need to use the apt cookbook anymore.


omnibus/cookbooks/chef-workstation-builder/recipes/default.rb
```diff
@@ -1,7 +1,7 @@
-# ensure packages available up-to-date
+# ensure packages are available and up-to-date
```
afiune: 😅

Issue #497, opened 2019-09-16
[PROPOSAL] [chef-analyze] Create a new CLI tool that leverages Chef Server API


afiune:
>As a Chef developer,
I would like to know if we can use the Golang to write libraries and binaries
that interact with the Chef Ecosystem, so we have the flexibility to choose
another language other than Ruby and speed up our CLI tools.

The **SPIKE** work has been completed and we are going to convert this Issue to be
the one that creates the new CLI tool that leverages Chef Server API.

# New Motivation
As a Chef user that wishes to know the current state of my infrastructure,
I would like to have a CLI tool that interacts with my Chef Server and analyzes
the Chef Client version of all my nodes and tells me a summary of the effort
needed to upgrade to the latest Chef Client version so that I have a full picture
of the steps I need to follow to be fully up-to-date.

## Description
At Chef, we would like to spike and understand how difficult would be to start
developing new binaries and/or libraries in Golang that interact with the Chef
Ecosystem.

Currently, there is already a repo from the community that abstracts the Chef
Server API client in Golang http://godoc.org/github.com/go-chef/chef, the first
goal of this spike is to understand this go library and its usability.

**Work with will done in the following repository: https://github.com/chef/chef-analyze**

## What problem(s) are we solving?

- **Speed**: All the Ruby CLIs we have at Chef are already in a state of considerable
slowness, things like `chef -v` or just a simple `knife help` takes 10-15m to run, let
alone commands that actually do something for the user, for example, `chef export policyfile`
or `knife node list`.
    - How can we make this experience better?
    - How can we run our CLIs faster?
- **Flexibility**: There is so much we can do in Ruby land, there is also a limit on how
much can we speed up our CLIs in that language, historically at Chef we have chosen
Ruby as our main language since the core of Chef (the product) is written in Ruby and
we would like to take advantage of the core code and behavior, though, there have been
cases where we don't need to share code and therefore, no need to use the same
language, such cases could be more efficient if we could/would use a compiled language,
but we haven't explored that path just yet. Having the ability to write our CLIs in any
language will allow us to have much more flexibility to improve all our tooling.
- **Organization**:  Another problem we could solve is the fact that, at Chef, we have
spread very wide in our toolings. We create multiple CLIs, commands, etc. that we deliver
to users in various manners, styles, and distributions, some of them are hard to identify
as a Chef tool. This work could improve the way we ship our tools and how users find
them in their system by having a single top-level command that has a list/catalog of sub-
[...]

PR #519, opened 2019-09-21
Top-level chef CMD wrapper

afiune:
In preparation for https://github.com/chef/chef-workstation/issues/497

# Design for a Top-Leve `chef` command

This top-level command will give us full control of our tooling inside the
Chef ecosystem, we would be able to integrate and modularize multiple
tools in a single place and make the `chef` command a first citizen of our
development experience.

A few usable examples of this design are:

1) On the creation of new CLIs, for instance, a new CLI called `chef-analyze` that
will provide some analyze capabilities, as a user, you can run the binary directly
`chef-analyze foo` or through our main Chef CLI wrapper `chef analyze foo`.
Both ways would be running the same underlying binary.

2) Redirecting existing binaries to a single point for further improments, this
could be the case of improving, for example, our `knife` tool that could take a few
seconds to run simple API searches like `knife node list`  or `knife cookbook list`

3) On replacements of current functionality, for example, if we would like to improve
the existing `chef generate foo` command, we would create a binary called
`chef-generate` that users can run individually like `chef-generate foo` and then
redirect the top-level CLI to run that same binary on any `chef generate XYZ`
execution.

Though, at the very beginning, we would pass every sub-command to the old `chef`
CLI binary that is being renamed to `chef-cli` with the following PR https://github.com/chef/chef-cli/pull/35,
being our default case it will give us backward compatibility with all commands and
a clean path for improvements.

## Related Resources
GH Issue: https://github.com/chef/chef-workstation/issues/540


afiune:
With this last commit I was able to build an omnibus package and install it on a brand new vm with the `chef-analyze` command available:
```
vagrant@default-ubuntu-1804:~/chef-workstation/omnibus$ sudo apt-get install -f /tmp/chef-workstation_0.9.42+20190926205134.git.8.aefe341-1_amd64.deb
Reading package lists... Done
Building dependency tree
Reading state information... Done
Note, selecting 'chef-workstation' instead of '/tmp/chef-workstation_0.9.42+20190926205134.git.8.aefe341-1_amd64.deb'
The following NEW packages will be installed:
  chef-workstation
0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.
Need to get 0 B/108 MB of archives.
After this operation, 335 MB of additional disk space will be used.
Get:1 /tmp/chef-workstation_0.9.42+20190926205134.git.8.aefe341-1_amd64.deb chef-workstation amd64 0.9.42+20190926205134.git.8.aefe341-1 [108 MB]
Selecting previously unselected package chef-workstation.
(Reading database ... 42487 files and directories currently installed.)
Preparing to unpack .../chef-workstation_0.9.42+20190926205134.git.8.aefe341-1_amd64.deb ...
Unpacking chef-workstation (0.9.42+20190926205134.git.8.aefe341-1) ...
Setting up chef-workstation (0.9.42+20190926205134.git.8.aefe341-1) ...
ldd: /opt/chef-workstation/components/chef-workstation-app/chef-workstation-app: No such file or directory

The experimental Chef Workstation App is available for you to try.

Launch the App by running 'chef-workstation-app'.
The App will then be available in the system tray.

Thank you for installing Chef Workstation!
You can find some tips on getting started at https://chef.sh/

vagrant@default-ubuntu-1804:~/chef-workstation/omnibus$ /opt/chef-workstation/bin/chef-analyze
One or more parameters missing.

Required:
	-user USER
	-key KEY
	-chef_server_url URL
```

tenor-250055883

components/main-chef-wrapper/main.go
```diff
@@ -0,0 +1,119 @@
...
+    install                 Install cookbooks from a Policyfile and generate a locked cookbook set
+    update                  Updates a Policyfile.lock.json with latest run_list and cookbooks
+    push                    Push a local policy lock to a policy group on the Chef Infra Server
+    push-archive            Push a policy archive to a policy group on the Chef Infra Server
+    show-policy             Show policyfile objects on the Chef Infra Server
+    diff                    Generate an itemized diff of two Policyfile lock documents
+    export                  Export a policy lock as a Chef Infra Zero code repo
+    clean-policy-revisions  Delete unused policy revisions on the Chef Infra Server
+    clean-policy-cookbooks  Delete unused policyfile cookbooks on the Chef Infra Server
+    delete-policy-group     Delete a policy group on the Chef Infra Server
```
tas50: An interesting experiment in the golang world would be setting up all these policy commands to exist under a "policy" group, while still providing backwards compatibility. This would let us clear up the tire fire of a a flat namespace we have while still allowing the existing commands to work:

``chef policy install`` would be the same as the legacy ``chef install`` from the chef-cli ruby gem.

afiune: Absolutely! If this gets in, we will have full control over a number, if not all our tools in the Chef Ecosystem. 

To that front, we need to have a good design to do such deprecations, new features, and more thought this main top-level wrapper.

afiune:
I am going to update this PR after a few other ones are merged! 💯

components/main-chef-wrapper/main.go
```diff
@@ -0,0 +1,123 @@
...
+		allArgs    = os.Args[2:]
+		cmd        *exec.Cmd
+	)
+
+	switch subCommand {
+	// 1) On the creation of new CLIs, for instance, a new CLI called chef-analyze that
+	// will provide some analyze capabilities, as a user, you can run the binary directly
+	// 'chef-analyze foo' or through our main Chef CLI wrapper 'chef analyze foo'.
+	// Both ways would be running the same underlying binary.
+	case "analyze":
```
afiune: I left this bit since we are going to merge this PR very soon: https://github.com/chef/chef-workstation/pull/553

marcparadise:
> Redirecting existing binaries to a single point for further improvements, this
could be the case of improving, for example, our knife tool that could take a few
seconds to run simple API searches like knife node list or knife cookbook list

This gives the potential to transition some commands at a more granular level too - for example  `knife search` can run a different a different binary than `knife bootstrap`.

afiune:
Required an adhoc build and last cleanup after the chef-analyze PR is merged.

Please, I would advocate leaving the commit history for our records since this was work originated from a Spike. cc/ @marcparadise @tyler-ball @jonsmorrow @tas50

components/main-chef-wrapper/main.go
```diff
@@ -0,0 +1,123 @@
...
+    describe-cookbook       Prints cookbook checksum information used for cookbook identifier
+
+Experimental Commands:
+    analyze                 Analyze your Chef inventory
+`)
+}
+
+func debugLog(msg string) {
+	if os.Getenv("DEBUG") == "true" {
+		fmt.Println(msg)
```
stevendanna: [nit] I usually find it helpful for debug logs to go to stderr.
```suggestion
		fmt.Fprintln(os.Stderr, msg)
```

components/main-chef-wrapper/main.go
```diff
@@ -0,0 +1,123 @@
...
+	// that users can run individually like chef-generate foo and then redirect the top-level
+	// CLI to run that same binary on any 'chef generate XYZ' execution.
+
+	// We want to pass every sub-command to the old 'chef' CLI binary that was renamed to
+	// 'chef-cli`, which is our default case.
+	default:
+		// When we land in the default case where we run the old 'chef' cli binary,
+		// we need to send the sub-command as well as all the arguments.
+		allArgs = append([]string{subCommand}, allArgs...)
+		cmd = exec.Command("chef-cli", allArgs...)
```
stevendanna: In the final implementation, I don't think you want your wrapper depending on the system PATH for this.  Perhaps we can hard-code the expected omnibus paths as constants, inject them at build-time, or use a relative path based on the location of this file.

components/main-chef-wrapper/main.go
```diff
@@ -0,0 +1,123 @@
...
+
+func main() {
+	// No arguments provided, display usage
+	if len(os.Args) <= 1 {
+		usage()
+		os.Exit(0)
+	}
+	var (
+		subCommand = os.Args[1]
+		allArgs    = os.Args[2:]
```
stevendanna: It might be worth it to spike an an early implementation of `-v` and `-h` just so you don't get too far down the road without figuring out how to handle partial option processing.  In some other wrappers, I've often resorted to just doing manual handling of the handful of top-level options since most libraries aren't good at parsing only a subset of options.

components/main-chef-wrapper/main.go
afiune: This is a good enhancement @stevendanna - I will make sure to add an Issue to work on this!

afiune: https://github.com/chef/chef-workstation/issues/562

components/main-chef-wrapper/main.go
afiune: @stevendanna do you mean to catch when the user provides either `chef -v` or `chef -h` commands? 🤔

afiune:
Built Passed!!!!! 

https://buildkite.com/chef/chef-chef-workstation-master-omnibus-adhoc/builds/111

tenor-112384116

components/main-chef-wrapper/main.go
afiune: I see your +1 @stevendanna - It makes sense to spike on that implementation, especially since this top-level command will allow us to standardize a lot of those flags and or config details, though, we would like to be as transparent as possible in this first iteration to the end-user, have feedback about the new integration with the new tools we will create and see if it makes sense to start implementing and improving existing functionality. 

Do you see any other blocker for us to merge and continue iterating on this approach/idea?

PR #556, opened 2019-10-02
[pipeline-fix] Keep the `chef` binary around until the new one is ready.

marcparadise:
This fixes the broken build. 

We validate that `chef` top-level command works in the
omnibus-test.* script. These tests explicitly look for the
command in the omnibus bin dir, which does not have a `chef` command anymore -
only chef-cli.

## Related Issues

introduced with #551 

## Types of changes
 (label applied)


afiune:
This PR is not needed since we have merged https://github.com/chef/chef-workstation/pull/519

PR #644, opened 2019-10-31
Avoid shipping double rdoc / minitest gems

tas50:
Pin minitest and 5.11.3 which is the version now being brought in and
avoid the rdoc pin entirely since that seems to be causing a double
verion.


tas50:
I confirmed this works in an adhoc and drops our builds by 2.9 megs on disk and 249 files.

Issue #640, opened 2019-10-31
Strip unnecessary files from the gems we ship in Workstation


tas50:
Things we have direct access to:

- [ ] appbundler
- [ ] berkshelf
- [ ] chef-apply
- [ ] chef-bin
- [ ] chef-cli
- [ ] chef-zero
- [ ] cheffish
- [ ] chefspec
- [ ] dep-selector-libgecode
- [ ] ffi-yajl
- [ ] knife-acl
- [x] knife-vcenter
- [ ] knife-windows
- [ ] libyajl2
- [x] license-acceptance
- [x] mdl
- [x] mixlib-authentication
- [ ] mixlib-install
- [ ] opscode-pushy-client
- [x] train-aws
- [x] train-habitat

Other gems where we'll need to open a PR

- [x] addressable
- [ ] akami
- [ ] ast
- [ ] backports
- [ ] bcrypt_pbkdf
- [ ] binding_of_caller
- [ ] builder
- [ ] citrus
- [ ] cleanroom
- [ ] coercible
- [ ] cucumber-tag_expressions
- [ ] debug_inspector
- [ ] declarative
- [ ] declarative-option
- [ ] descendants_tracker
- [ ] did_you_mean
- [ ] diff-lcs
- [ ] domain_name
- [ ] ed25519
- [ ] equalizer
- [ ] equatable
- [ ] erubis
- [ ] faraday-cookie_jar
- [ ] ffi
- [ ] ffi-rzmq
- [ ] ffi-rzmq-core
- [ ] filesize
- [ ] formatador
- [ ] gcewinpass
- [ ] gherkin
- [ ] google-api-client
- [ ] google-protobuf
- [ ] googleauth
- [ ] gssapi
- [ ] gyoku
- [ ] hashie
- [ ] highline
- [ ] htmlentities
- [ ] http_parser.rb
- [ ] http
- [ ] http-cookie
- [ ] http-form_data
- [ ] httpclient
- [ ] httpi
- [ ] ice_nine
- [ ] inifile
- [ ] iniparse
- [ ] ipaddress
- [ ] json
- [ ] jwt
- [ ] kartograph
- [ ] kramdown
- [ ] kramdown-parser-gfm
- [ ] little-plugger
- [ ] lockfile
- [ ] logging
- [ ] lumberjack
- [ ] memoist
- [ ] method_source
- [ ] mime-types
- [ ] mime-types-data
- [ ] mini_portile2
- [ ] minitar
[...]

PR #672, opened 2019-11-04
Aggressively prune files from the gems we install

tas50:
See what we can do about our overall file size and count by removing
non-library files from the package.

 512MB - > 493MB. 3095 files removed

### DIFF
--- a/omnibus/config/software/more-ruby-cleanup.rb
+++ b/omnibus/config/software/more-ruby-cleanup.rb
@@ -1,0 +1,104 @@
+#
+# Copyright:: 2019 Chef Software, Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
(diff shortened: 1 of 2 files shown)


tas50:
Paging @lamont-granquist since he probably has opinions here as well

tas50:
> Is there a way we can install the gems without all these files instead of removing the files as part of a clean-up-process?

Nope. We can only open PRs for each of the 100 gems to remove the files, but so far about 50% have said no since they want all their test files to ship in the gems.

omnibus/config/software/more-ruby-cleanup.rb
```diff
@@ -0,0 +1,104 @@
...
+      CHANGELOG.txt
+      warning.txt
+      FAQ.txt
+      release-script.txt
+      TODO
+      HISTORY
+      CHANGES
+      CHANGELOG
+      README
+      Gemfile
```
lamont-granquist: think you broke appbundler here

omnibus/config/software/more-ruby-cleanup.rb
```diff
@@ -0,0 +1,104 @@
...
+      README.rdoc
+      CHANGELOG.rdoc
+      History.rdoc
+      CONTRIBUTING.rdoc
+      README_INDEX.rdoc
+      logo.png
+      donate.png
+      Appraisals
+      website
+      *.gemspec
```
lamont-granquist: and probably here

lamont-granquist:
seems fine other than those two

omnibus/config/software/more-ruby-cleanup.rb
tas50: This is at the end once appbundler has already done its thing. Is that going to be a problem still?

tas50:
@afiune this is the list of what we'd need to cleanup at the gem level to avoid this kind of built time cleanup: https://github.com/chef/chef-workstation/issues/640

PR #714, opened 2019-11-14
Cleanup additional files / folders in the installed gems

tas50:
This slims the install by 241k on disk and 28 files.


tas50:
Cleaning the tests is the next thing and that's going to have to be a bit more surgical

PR #728, opened 2019-11-19
Update Chef Infra Client to 15.5.15

tas50:
Also add a sleep to the update script so this works automatically next
time


tas50:
This is handled in #728

PR #817, opened 2019-12-20
More improvements to the Windows installer

tas50:
1) We were never enabling Chef Workstation App. We were installing it. This updated wording is consistent with the actual Chef Workstation wording
2) Rename some components so it's more clear to folks in the future what they do
3) Call it the Chef Workstation Taskbar App in the description so it's more clear what you're selecting
4) Enable the WS App subfeature to enable it at boot. Have this be disabled by default, but this lets us at least test this functionality for when (soon) we make it default


tas50:
Here's the view when both features have been expanded:

image

PR #877, opened 2020-01-23
(design_proposal): chef cli catalog

afiune:
## Description
A design proposal for the future of the `chef` CLI Catalog.

## Motivation
```
As a Chef Operator,
I need a unified Command Line Interface (CLI) that lets me interact with every Chef product,
so I can easly discover the capabilities of the Chef ecosystem.

As a Chef Developer,
I need to be able to gather information about how our users consume and interact with our tools,
so I can identify which commands are used the most and measure the usability and performance on all supported platforms.
```
## Acceptance Criteria

- [ ] A design doc is stored inside the code repo
- [ ] The design doc has been approved by everyone on the team
- [x] Design doc has enough details / shared understanding that any engineer could write correctly scoped cards based on it
- [x] Update the linked epic card to capture the scope of this design

 

docs/Chef-CLI-Catalog.md
```diff
@@ -0,0 +1,91 @@
+# Chef CLI (Catalog)
+
+Throughout the history of Chef, we have created and acquired a variety
+of tools that have grown to a point that, as a Chef practitioner, it is
```
jonsmorrow: Chef Operator to bring this in-line with our current target user base.

docs/Chef-CLI-Catalog.md
```diff
@@ -0,0 +1,91 @@
...
+* delivery
+* chef-apply
+* chef-client
+* chef-run
+* chef-shell
+* chef-solo
+
+## Goals
+* Have a single, unified way to discover tools to interact with Chef products
+* Huge speed improvements
```
jonsmorrow: All of the other goals are more tangible. I'd say something like: "Consistent/Usable performance on all supported platforms"

docs/Chef-CLI-Catalog.md
```diff
@@ -0,0 +1,91 @@
...
+    As a Chef practitioner,
+    I would like to have a single Command Line Interpreter (CLI) that lets me easily interact with every Chef product.
+    I would also like to have a way to discover new tools that can help me with my day to day work.
+    Finally, since I run these tools daily, I need them to run fast
+
+    As a Chef developer,
+    I would have a tremendous benefit from this architecture since
+    I would be able to gather information about how our users consume and interact with our tools,
+    as well as identifying the commands that are used the most,
+    so I can improve the speed and/or do extreme modifications to the code base (do a rewrite in a different, more efficient, programing language)
```
jonsmorrow: This so states things you could do not why you'd want to. Is there a way to say this without talking about extreme changes. Anytime we use that kind of language it can alienate folks that want more measured changed/no change. It's ok if you really mean to drive extreme change and understand the above, but make sure that's the intent.

docs/Chef-CLI-Catalog.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Motivation
+
+    As a Chef practitioner,
+    I would like to have a single Command Line Interpreter (CLI) that lets me easily interact with every Chef product.
+    I would also like to have a way to discover new tools that can help me with my day to day work.
+    Finally, since I run these tools daily, I need them to run fast
+
+    As a Chef developer,
+    I would have a tremendous benefit from this architecture since
```
jonsmorrow: These I woulds are future tense assuming this has been implemented. I think they need to be written as what is desired. It's also pre-judging the architecture before garnering feedback. I'd suggest more along the lines of: I would like a more flexible architecture that allows me to invest in improvements selectively where they will have the most impact.

docs/Chef-CLI-Catalog.md
```diff
@@ -0,0 +1,91 @@
...
+    I would be able to gather information about how our users consume and interact with our tools,
+    as well as identifying the commands that are used the most,
+    so I can improve the speed and/or do extreme modifications to the code base (do a rewrite in a different, more efficient, programing language)
+
+## Specification
+
+Currently, inside the latest version of Chef Workstation (0.14.16 as of 01/20/2020),
```
jonsmorrow: It's been there longer, so I'd find when it landed or just remove the version.

docs/Chef-CLI-Catalog.md
```diff
@@ -0,0 +1,91 @@
...
+the team has implemented a top-level chef command that acts as a wrapper around our
```
jonsmorrow: s/chef/`chef`

docs/Chef-CLI-Catalog.md
```diff
@@ -0,0 +1,91 @@
...
+variety of tools. The main idea about this command is to have the ability to create
```
jonsmorrow: wrapper around chef commands.

jonsmorrow: Here you are future tense again as if it's been done. I think you want to say something like this design proposes we expand this scope to be a proxy/catalog of all the user facing cli tools we ship.

docs/Chef-CLI-Catalog.md
```diff
@@ -0,0 +1,91 @@
...
+a catalog of tools/sub-commands, it will act as a proxy between the user and our
+tools.
+
+Quick diagram that illustrates the new Chef CLI Catalog:
+
+![chef-cli-catalog](img/chef-top-level-command.jpg)
+
+With this architecture, we are now able to integrate and modularize multiple tools
```
jonsmorrow: s/we are now able/we would be able

docs/Chef-CLI-Catalog.md
```diff
@@ -0,0 +1,91 @@
...
+in a single place and make the chef command a first citizen of our development
+experience. We will also be able to gather immediate information from our users of
+all our tools so we can understand what parts of our tooling are the most used and
+which areas can we improve.
+
+Once detecting such areas of improvement, we can have the flexibility to improve
+tools entirely or sub-sections of the tool. Say for instance, that our users run
+`knife search` around 30-50 times a day, if the command takes around 10-15
+seconds to run, we are consuming ~10 minutes from our users only on waiting time.
```
jonsmorrow: ~10 minutes a day from our users waiting for value.

docs/Chef-CLI-Catalog.md
```diff
@@ -0,0 +1,91 @@
...
+[Gist with a few mocked help commands.](https://gist.github.com/afiune/1dc854089002e182288a0452eaa91908)
+
+## Downstream Impact
+TBA
+
+## Milestones
+### Implement Telemetry into the top-level chef command
+We need to start gathering information about how our users use the chef CLI,
+this will allow us to understand the impact that we will have from the
+aggressivemoves we are about to do.
```
jonsmorrow: Can we do this in a non-aggressive way?

docs/Chef-CLI-Catalog.md
```diff
@@ -0,0 +1,91 @@
...
+
+![mocked-chef-help-command](img/mocked-chef-help-command.png)
+
+[Gist with a few mocked help commands.](https://gist.github.com/afiune/1dc854089002e182288a0452eaa91908)
+
+## Downstream Impact
+TBA
+
+## Milestones
+### Implement Telemetry into the top-level chef command
```
jyan-chef: Just FYI, there's some dependency here on business requirements and legal requirements. Essentially we need to solidify our business requirements, get buy in from potential stakeholders (i.e. outside of Prod/Eng) and then get legal approval to proceed. This work is essential but just want to call out that the requirements are a little TBD right now and we have legal hurdles to overcome.

docs/Chef-CLI-Catalog.md
```diff
@@ -0,0 +1,91 @@
+# Chef CLI (Catalog)
+
+Throughout the history of Chef, we have created and acquired a variety
+of tools that have grown to a point that, as a Chef practitioner, it is
+hard to know, discover and understand all of them. A few of the tools
+that our users use (almost) every day are:
```
jyan-chef: I think having a catalog of all available commands/tools is very important. Once the user becomes aware of a tool, the next step is really for them to figure out what it does and if it's useful to them. Is that part of the intended scope of this work? If so then we can collaborate w/ UX team on figuring where/how to display relevant information about each tool and maybe include URLs to more detailed information (e.g. docs page) to users

docs/Chef-CLI-Catalog.md
afiune: yup, since `0.10.41`

jonsmorrow:
Jon will take point on driving this with Jessica and coordinating with salim for any updates to the design.

docs/Chef-CLI-Catalog.md
afiune: Yes, this is exactly one of the ideas/goals and therefore, collaboration with UX will be required.

docs/Chef-CLI-Catalog.md
```diff
@@ -0,0 +1,228 @@
...
+can control the "Usage" message of sub-binaries. Currently, the `chef-analyze`
+binary displays the usage message `chef-analyze [command]` when it is run directly
+and also when it is run from the `chef` catalog. We would like the usage message
+to be consistent with the way users execute the commands so if you run `chef analyze`
+the usage sould be `chef analyze [command]`; and if you run `chef-analyze` the
+usage should continue to be `chef-analyze [command]`.
+
+## Open points for further discussion:
+* Are we discouraging the use of individual sub-binaries like `knife` or `chef-run`?
+* What happens when a user runs a sub-binary without the prefix `chef`?
```
tyler-ball: IMO, we should not communicate the existence of these binaries to users but we should not prevent their usage. It is just unsupported usage.

docs/Chef-CLI-Catalog.md
```diff
@@ -0,0 +1,228 @@
...
+    shell-init              Initialize your shell to use Chef Workstation as your primary Ruby
+    generate                Generate a new repository, cookbook, or other component
+    describe-cookbook       Prints cookbook checksum information used for cookbook identifier
+    policy                  Manage Chef Policyfiles
+
+Relocated Commands:
+    install                 Use now: chef policy install
+    update                  Use now: chef policy update
+    push                    Use now: chef policy push
+    push-archive            Use now: chef policy push-archive
```
tyler-ball: Going with the CLI UX `noun verb` guidelines I think the following make more sense:

```
chef policy install
chef policy update
chef policy push
chef policy show
chef policy diff
chef policy export
chef policy delete
chef policy undelete
chef policy-archive push
chef policy-cookbooks clean
chef policy-group delete
chef policy-revisions clean
```

Under the covers, `chef policy-group delete` could still invoke `chef-policy group delete` but that would be an internal implementation detail. It would make it easier for us to implement because it would be less changing of existing code. But I think the public contract is more important to cover in this design doc than the internal implementation of our command routing.

Thoughts?

docs/Chef-CLI-Catalog.md
tyler-ball: I misunderstood this line - I was thinking that we should document `chef policy` and not document `chef-policy` but if they run `chef-policy` we don't do anything to prevent them from doing that.

docs/Chef-CLI-Catalog.md
afiune: From our conversation over zoom we have agreed that these commands look correct with the exception of `chef-policy push-archive` that will be removed and include into the `chef-policy push` command.

tyler-ball: More context - I convinced myself that switching to a strictly enforced 'noun verb' syntax means we end up with a bunch of unnecessary nouns (`policy-revision`, `policy-cookbooks`, `policy-groups`, etc.) with 1 verb each.

Instead we talked over some of the existing commands and think the PR correctly documents them. The 'noun' in this case is always 'policy' because all these commands are all connected to Policyfiles.

The only tricky one is `chef-policy clean cookbooks` because it deletes cookbooks that are not referenced by a policyfile. Maybe this becomes an alias for something like `chef cookbooks clean`?

docs/Chef-CLI-Catalog.md
```diff
@@ -0,0 +1,226 @@
...
+    delete-policy           Use now: chef policy delete
+    undelete                Use now: chef policy undelete
+
+Experimental Commands:
+    analyze                 Analyze artifacts from a Chef Infra Server
+```
+
+### EXTRA: Present correct usage of sub-binaries
+
+An additional task from this proposal is the use of an environment variable that
```
tyler-ball: Talked with Salim about this - IMO, our first implementation should reference `chef policy` in all the help text. This makes it more confusing for our effortless developers who will directly invoke the `chef-policy` binary (without all the overhead of the entire Chef Workstation ecosystem). But those are internal users we can help.

In the future I could see us developing a Golang `effortless` executable that functions similarly to the `chef` golang executable. It allows us to start creating a custom workflow for them, leveraging tools under the covers that they do not care whether are written in Ruby or Golang. At that time I would introduce some kind of environment variable that would allow the help text to display `chef policy` or `effortless policy`.

docs/Chef-CLI-Catalog.md
```diff
@@ -0,0 +1,226 @@
...
+can control the "Usage" message of sub-binaries. Currently, the `chef-analyze`
+binary displays the usage message `chef-analyze [command]` when it is run directly
+and also when it is run from the `chef` catalog. We would like the usage message
+to be consistent with the way users execute the commands so if you run `chef analyze`
+the usage sould be `chef analyze [command]`; and if you run `chef-analyze` the
+usage should continue to be `chef-analyze [command]`.
+
+## Open points for further discussion:
+* Are we discouraging the use of individual sub-binaries like `knife` or `chef-run`?
+* What happens when a user runs a sub-binary without the prefix `chef`?
```
tyler-ball:
```suggestion
* What happens when a user runs a sub-binary without the prefix `chef`?
    * How do we capture telemetry on usage when users directly access tools? (EG, `berks`, `inspec` instead of `chef berks`, `chef inspec`)
```

tyler-ball:
Notes from group discussion: We want to add "auto-generated docs for tools/products" to the goals for this proposal.

docs/Chef-CLI-Catalog.md
```diff
@@ -0,0 +1,226 @@
...
+Available Commands:
+    env                     Prints environment variables used by Chef Workstation
+    gem                     Runs the 'gem' command in context of the embedded Ruby
+    exec                    Runs the command in context of the embedded ruby
+    shell-init              Initialize your shell to use Chef Workstation as your primary Ruby
+    generate                Generate a new repository, cookbook, or other component
+    describe-cookbook       Prints cookbook checksum information used for cookbook identifier
+    policy                  Manage Chef Policyfiles
+
+Relocated Commands:
```
tyler-ball:
```suggestion
Renamed Commands:
```

docs/Chef-CLI-Catalog.md
```diff
@@ -0,0 +1,226 @@
...
+    delete-policy           Use now: chef policy delete
+    undelete                Use now: chef policy undelete
+
+Experimental Commands:
+    analyze                 Analyze artifacts from a Chef Infra Server
+```
+
+#### Open questions for further discussion with UX:
+* How do we want to present these sections? What is the right order?
+* What verbiage should we use to express the relotation of commands? (`Use now` Vs `Moved to`)
```
tyler-ball: `now` is unnecessary, can just say `Use`

docs/Chef-CLI-Catalog.md
```diff
@@ -204,6 +204,19 @@ Experimental Commands:
     analyze                 Analyze artifacts from a Chef Infra Server
 ```
 
+### Automation of reference documentation
+
+Motivation:
+
+    As a technical writer
+    I want to add examples to the chef catalog documentation
+    so that users will know the tricky bits and we the documentations gets updated automatically as the code evolves
```
afiune:
```suggestion
    so that users will know the tricky bits and the documentation gets updated automatically as the code evolves
```

docs/Chef-CLI-Catalog.md
jyan-chef: is it correct to assume the wrapper has been applied to all commands for all binaries included in workstation?

docs/Chef-CLI-Catalog.md
jyan-chef: Can someone help me understand how catalog improvements are links to performance? (i.e. how does the catalog work speeds things up)

docs/Chef-CLI-Catalog.md
jyan-chef: With regard to telemetry, how can we link this to legal requirements around data capture so the design doc includes the our internal privacy requirements? (e.g. how can we log requirements from legal such as only turn on anonymous data capture after accepting the EULA?)

afiune:
tenor-44698470

PR #885, opened 2020-01-24
Update InSpec to 4.18.51 and pin train to 3.2.0

tas50:
We're still seeing an issue with the latest train and kitchne-inspec so we should continue to pin this.


tas50:
The pin is due to https://github.com/inspec/train/issues/548

PR #886, opened 2020-01-24
Exclude omnibus_packages in chef bundle testing

tas50:
Due to pins in inspec vs. inspec-core in 4.18.51 we're getting failures.
This will be resolved with a new inspec release, but that doesn't
unblock us today.

### DIFF
--- a/habitat/plan.sh
+++ b/habitat/plan.sh
@@ -112,7 +112,7 @@ do_install() {
     wrap_ruby_bin "chef-resource-inspector"
     wrap_ruby_bin "chef-shell"
 
-    appbundle "chef" "docgen,chefstyle"
+    appbundle "chef" "docgen,chefstyle,omnibus_packages"
     wrap_ruby_bin "knife"
 
     appbundle "inspec-bin" "changelog,debug,docs,development"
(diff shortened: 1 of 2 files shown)


lamont-granquist:
For posterity:

This affects the transitive merge of all of chef-workstation's Gemfile.lock "runtime" deps with the chef gem's "development" gems (really all the gems that aren't in the gemspec, not just the literal development group).

The without argument does not know anything about the chef-workstations gem groups in the Gemfile.lock, it just slurps up the whole thing, doesn't care.

What it applies to is the chef gems Gemfile, and omits groups there.

It does not affect shipping gems.  So the fact that both Gemfile's have an "omnibus_packages" group is confusing, but this does not apply to the chef-workstation side and that group has already been installed in the bundle install.

In this case the "omnibus_packages" group in the chef Gemfile had conflicting transitive gems.  None of those gems are necessary for running rspec on the testers to test the chef gem against the workstation gemset.  This is an example of exactly why the "without" flag was built -- builds can be unblocked simply by removing the unnecessary gems rather than having to reconcile dev deps across the entire chef-workstation ecosystem globally.

If anyone gets distracted by noticing that in an ideal world one would want to "DRY" the duplication in gems between the omnibus_packages groups in both applications -- stop think about that -- Gemfile's themselves aren't ever transitive due to bundler and that will never happen and can't happen, and we cannot and will not manufacture that kind of feature, so its a very distracting mental rathole.   That isn't happening, can't happen, and isn't what this is about.

Really every gem group in the chef/Gemfile.lock which is not strictly required for doing rspec testing should be included in the without line.

PR #924, opened 2020-02-07
Create the .chef dir automatically on mac/linux/windows

tas50:
This is one of the manual things you have to do post Workstation install that seems entirely silly. We should just create this directory for them automatically so we can remove this step from the docs.

### DIFF
--- a/omnibus/package-scripts/chef-workstation/postinst
+++ b/omnibus/package-scripts/chef-workstation/postinst
@@ -83,7 +83,7 @@ for binary in $binaries; do
 done
 
 # everyone needs a .chef dir so let's create it for them
-if [ ! -d "~/.chef/" ]; then
+if [ ! -d ~/.chef/ ]; then
   mkdir ~/.chef
   chmod 700 ~/.chef
 fi
--- a/omnibus/package-scripts/chef-workstation/postinst
+++ b/omnibus/package-scripts/chef-workstation/postinst
@@ -83,7 +83,7 @@ for binary in $binaries; do
 done
 
 # everyone needs a .chef dir so let's create it for them
-if [ ! -d ~/.chef/ ]; then
+if [ ! -d "~/.chef/" ]; then
   mkdir ~/.chef
   chmod 700 ~/.chef
 fi
--- a/omnibus/package-scripts/chef-workstation/postinst
+++ b/omnibus/package-scripts/chef-workstation/postinst
@@ -82,6 +82,12 @@ for binary in $binaries; do
   ln -sf "$INSTALLER_DIR/bin/$binary" $PREFIX/bin || error_exit "Cannot link $binary to $PREFIX/bin"
 done
 
+# everyone needs a .chef dir so let's create it for them
+if [ ! -d ~/.chef/ ]; then
+  mkdir ~/.chef
+  chmod 700 ~/.chef
+fi
+
 if is_darwin; then
   # the app launcher comes from the chef-workstation-app repo, here we are just using
   # it to start the app as a service on MacOS systems, it will start at boot
--- a/omnibus/resources/chef-workstation/msi/assets/create-chef-dir.ps1
+++ b/omnibus/resources/chef-workstation/msi/assets/create-chef-dir.ps1
@@ -1,0 +1,6 @@
+$chef_path = "$HOME\.chef"
+
+If(!(test-path $chef_path))
+{
+  New-Item -ItemType Directory -Force -Path $chef_path
+}
--- a/omnibus/resources/chef-workstation/msi/assets/create-chef-dir.ps1
+++ b/omnibus/resources/chef-workstation/msi/assets/create-chef-dir.ps1
@@ -2,5 +2,5 @@ $chef_path = "$HOME\.chef"
 
 If(!(test-path $chef_path))
 {
-  New-Item -ItemType Directory -Force -Path $chef_path
+  New-Item -ItemType Directory -Path $chef_path
 }
--- a/omnibus/resources/chef-workstation/msi/assets/create-chef-dir.ps1
+++ b/omnibus/resources/chef-workstation/msi/assets/create-chef-dir.ps1
@@ -1,4 +1,4 @@
-$chef_path = "$HOME\.chef"
+$chef_path = "$env:USERPROFILE" + "\.chef"
 
 If(!(test-path $chef_path))
 {
--- a/omnibus/resources/chef-workstation/msi/source.wxs.erb
+++ b/omnibus/resources/chef-workstation/msi/source.wxs.erb
@@ -94,32 +94,32 @@
        For now, using COMMON_ALTSTARTUP which is the global auto-start folder. Note that this
        may cause some trouble - we haven't yet looked into how the App should behave when multiple
        instances are running across different users.
-  -->
-  <PropertyRef Id="WIX_DIR_COMMON_ALTSTARTUP" />
+    -->
+    <PropertyRef Id="WIX_DIR_COMMON_ALTSTARTUP" />
 
-  <CustomActionRef Id="WixBroadcastSettingChange" />
-  <CustomActionRef Id="WixBroadcastEnvironmentChange" />
+    <CustomActionRef Id="WixBroadcastSettingChange" />
+    <CustomActionRef Id="WixBroadcastEnvironmentChange" />
 
-  <SetDirectory Id="WINDOWSVOLUME" Value="[WindowsVolume]"/>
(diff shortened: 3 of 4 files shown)


marcparadise:
Since the install will often be done via sudo, won't the home dir in question be /root/.chef? 

I'm wondering if this wouldn't be better done inside of the `chef` bin itself - I think it's the only one that fails in absence of `chef`, and it makes sense it would create it if it's not present.

tyler-ball:
@marcparadise I just checked and `sudo ls ~/` showed the home directory for _my_ user on MacOS. We can check this via an ad-hoc build though

marcparadise:
@tyler-ball it may depend on the sudo flags/defaults

tas50:
Sudo shouldn't matter:

A simple script that echos ~

```
❰tsmith❙~❱✔≻ ./tim.sh
/Users/tsmith
❰tsmith❙~❱✔≻ sudo ./tim.sh
Password:
/Users/tsmith
```

tas50:
If someone isn't passing the env vars to sudo I'd considered that a solid edge case and they can create the dir themselves. This would solve it for 99% of folks. Doing it in chef wouldn't fix it for all the knife users who wouldn't have the dir on the first run either.

marcparadise:
Cool, I thought knife already handled it since it I'd only seen this causing an error in `chef/chef-analyze`.

tas50:
```
❰tsmith❙~❱✔≻ mv .chef .chef_backup
❰tsmith❙~❱✔≻ knife node list
WARNING: No knife configuration file found. See https://docs.chef.io/config_rb_knife.html for details.
WARN: Failed to read the private key /etc/chef/client.pem: #<Errno::EACCES: Permission denied @ rb_sysopen - /etc/chef/client.pem>
ERROR: Your private key could not be loaded from /etc/chef/client.pem
Check your configuration file and ensure that your private key is readable
❰tsmith❙~❱✘≻ ls -lah .chef
ls: .chef: No such file or directory
```

omnibus/resources/chef-workstation/msi/assets/create-chef-dir.ps1
```diff
@@ -0,0 +1,6 @@
+$chef_path = "$HOME\.chef"
+
+If(!(test-path $chef_path))
+{
+  New-Item -ItemType Directory -Force -Path $chef_path
+}
```
afiune: missing newline

omnibus/resources/chef-workstation/msi/assets/create-chef-dir.ps1
```diff
@@ -0,0 +1,6 @@
+$chef_path = "$HOME\.chef"
```
afiune: Does `$HOME` work on windows? I thought it was `$env:USERPROFILE` instead 🤔

omnibus/resources/chef-workstation/msi/assets/create-chef-dir.ps1
```diff
@@ -0,0 +1,6 @@
...
+
+If(!(test-path $chef_path))
+{
+  New-Item -ItemType Directory -Force -Path $chef_path
```
afiune: Why do you force this action? `-Force`

omnibus/package-scripts/chef-workstation/postinst
```diff
@@ -31,6 +31,12 @@ for binary in $binaries; do
   ln -sf "$INSTALLER_DIR/bin/$binary" $PREFIX/bin || error_exit "Cannot link $binary to $PREFIX/bin"
 done
 
+# everyone needs a .chef dir so let's create it for them
+if [ ! -d ~/.chef/ ]; then
```
afiune: It is recommended to use double quotes to avoid problems during string expansion,
`"~/.chef"` and when you do that, you can't use `~` so I would recommend the use
of `$HOME` here and in the next two lines.

omnibus/resources/chef-workstation/msi/assets/create-chef-dir.ps1
tas50: to prevent it from blowing up if that's already there

tas50: which actually shouldn't be necessary anymore

tyler-ball:
This does not solve the issue in a multi-user workstation, but we don't think that is an often seen case. Moving this to the top of our backlog to get it prioritized for finishing and merging.

omnibus/package-scripts/chef-workstation/postinst
tas50: Fixed

omnibus/resources/chef-workstation/msi/assets/create-chef-dir.ps1
tas50: Fixed

omnibus/resources/chef-workstation/msi/source.wxs.erb
```diff
@@ -94,32 +94,37 @@
...
+                <Environment Id="Environment"
+                  Name="PATH" Action="set" Part="last" System="yes" Value="[PROJECTLOCATIONBIN]" />
+              </Component>
+              <Component Id="ChefWSRegistryEntries" Guid="{67E30C25-5311-4584-9A17-5D47AC28F9CD}">
+                <RegistryKey Root="HKLM" Key="Software\Chef\Chef Workstation">
+                  <RegistryValue Type="string" Name="InstallDir" Value="[PROJECTLOCATION]" />
+                  <RegistryValue Type="string" Name="BinDir" Value="[PROJECTLOCATIONBIN]" />
+                </RegistryKey>
+              </Component>
+              <Component Id="CreateDotChefDirPowershellScript" Guid="{DB58E387-1D7C-4D3D-8413-372208B82A8C}" >
```
marcparadise, line 119: Most of the diff here is spacing.  The actual change in this section is adding 'CreateDotChefDirPowershellScript' Component

marcparadise:
In testing this manually, I found that on Debian and RHEL (so far) the postinst script `~/.chef` does expand to `/root/.chef` when it's being run by th einstaller. 

That makes sense, and is consistent with @tyler-ball's finding above where ` sudo ls ~/` showed the directory of the current user - the `~` would be expanded by bash before anything wsa invoked, so it would be in the context of the current user and not `sudo`.  When the expansion occurs in-script, it's different because the script is already running as root. 

TL;DR: I've updated to use `$HOME` instead of `~`, running through pipeline to verify.

marcparadise:
More digging has shown that we can't rely on $HOME being correct either. I think we're going to have go a different route - the default configuration seems to be that $HOME will not be carried forward, based on the results seen on mac/centos/ubuntu.  

Example: 
```
[vagrant@ws-test ~]$ bash -c 'echo $HOME'
/home/vagrant
[vagrant@ws-test ~]$ sudo bash -c 'echo $HOME'
/root
[vagrant@ws-test ~]$ sudo --preserve-env bash -c 'echo $HOME'
/root
[vagrant@ws-test ~]$ sudo --preserve-env=HOME bash -c 'echo $HOME'
/home/vagrant
```

 On some platforms, `--preserve-env` or `-E` will default to including $HOME, but some will not; and in any case we don't want to rely on `sudo` being run with flags to ensure.  We'd also need `$USER` to ensure correct dir ownership. 

 I'm much less comfortable relying on sane sudo defaults, since this seems to be the default behavior across at least a few different distros.

tas50:
@marcparadise does the windows side of this work?

marcparadise:
Windows  is failing for reasons I haven't debugged yet - I need to turn up log level on the installer run. I do think we'll be able to get this working in the MSI one way or another (maybe as a component instead of a script); but I'm not sure if we want to do that given we need another approach anyway for *nix.

tyler-ball:
https://unix.stackexchange.com/questions/23463/how-to-create-debian-package-to-install-files-to-home-user
https://stackoverflow.com/questions/35971583/debian-package-creation-postinst-as-non-root

I am not experienced in this area, but some googling looks like what we're trying to do isn't necessarily how the *nix package managers were designed to work...

Maybe we put something in the `chef` Golang wrapper so it is at least faster?

marcparadise:
I think we'd talked about this a bit in one of the triage discussions.  *nix packages are not expected to touch user files. I don't think it's violating any spec to do so, but it seems avoided by convention. 

Using the `chef` wrapper matches my earlier thought as well.    It won't be quite as complete - non-wrapped commands won't benefit unless they're invoked via `chef execute`[1], but it will address it for `chef` and all its subcommands.  


1)  at least not unless/until we proceed with the binstub changes we'd been considering in the context of the Chef CLI Catalog.

PR #958, opened 2020-02-14
Set version of packages on separate line

bobchaos:
All packages that are typically replaced by 3rd party distros now have parameters like version on separate lines to simplify patching
Expeditor scripts that handle version bumps have been updated accordingly

This will allow 3rd party distros to patch in their own gems without having to generate new patches with each version bump.

Signed-off-by: Marc Chamberland <chamberland.marc@gmail.com>

## Description

See above

## Related Issue

none, went straight to PR!

## Types of changes

- [X] Chore (non-breaking change that does not add functionality or fix an issue)

## Checklist:

- [X] I have read the **CONTRIBUTING** document.
- [NA] I have run the pre-merge tests locally and they pass.
- [NA] I have updated the documentation accordingly. # I could add in-line comment if need be
- [NA] I have added tests to cover my changes. # no code change
- [NA] All new and existing tests passed. # no code change
- [X] All commits have been signed-off for [the Developer Certificate of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

Aha! Link: https://chef.aha.io/features/SH-1491

### DIFF
--- a/.expeditor/update_chef-analyze_to_latest.sh
+++ b/.expeditor/update_chef-analyze_to_latest.sh
@@ -16,7 +16,7 @@ version=$(get_github_file "$EXPEDITOR_REPO" master VERSION)
 branch="expeditor/chef-analyze${version}"
 git checkout -b "$branch"
 
-sed -i -r "s/override \"chef-analyze\",\s+version: \"[^\"]+\"/override \"chef-analyze\", version: \"${version}\"/" omnibus_overrides.rb
+sed -i -r "/^override \"chef-analyze\",/{\$!{N;s/version: \".*\"/\"= ${version}\"/}}" omnibus_overrides.rb
 
 git add .
 
--- a/.expeditor/update_chef-workstation-app_to_latest.sh
+++ b/.expeditor/update_chef-workstation-app_to_latest.sh
@@ -16,7 +16,7 @@ version=$(get_github_file $EXPEDITOR_REPO master VERSION)
 branch="expeditor/chef_workstation_app_${version}"
 git checkout -b "$branch"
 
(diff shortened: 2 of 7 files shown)


bobchaos:
The sed commands feel a bit brittle for multi-line setups; we assume the version is on the second line, and there's a distinct risk of modifying the version of the wrong gem should version constraints be removed in the future, ie:
```ruby
  gem "chef" # we just remove the version constraint for some reason
  gem "somegem", "= 1.1.1" # Now this line is matched! :O
```
I used the tools I found in the repo (sed), but if you can tell me what else is available on your expeditor executors I could build something sturdier, presumably.

tyler-ball:
We were looking at this with @ramereth during triage. He wanted to do a little testing to see if these commands would work.

I think reformatting our Gemfile to make it easier for the OSS community to modify is 💯 the right thing to do, so we will help during weekly triage to find the best way to do this. I don't know of other tools in our expeditor images that would be better for this than sed off the top of my head.

tduffield:
The other option is to use a `ruby` action. Expeditor can run `ruby` scripts the same way it can run `bash` scripts. So if you want to write something in Ruby, you could use that as well.

tyler-ball:
@ramereth @bobchaos I'm okay with using a sed command to handle this. We do not mess with these gems very often.

I think the one multi-line change we need to test is the `inspec` gems. Those are updated using the `update_dep.sh` script, which just wraps `bundle update` through a few layers of abstraction. Would you mind rebasing this PR and then testing that?

ramereth:
@tyler-ball apologies for the delay. I'll do my best to try testing this soon and verify

ramereth:
@bobchaos I don't think this is needed anymore, let's go ahead and close it.

PR #1061, opened 2020-03-25
Update Cookstyle to 6.3 / Chefstyle to 1.0

tas50:
This adds 7 new cops, adds autocorrection to a good number of cops, improves offense detection, and enables new built-in Rubocop cops. It also resolves several failures I reported in the Rubocop engine that caused cookstyle to fail when checking cookbooks on the supermarket. Chefstyle is bumped as well since the Rubocop engine versions need to match.


tyler-ball:
I rebased but I would like to wait to get this merged until I get the major version update merged. I'll continue working on that tomorrow

components/gems/Gemfile.lock
```diff
@@ -352,8 +352,8 @@ GEM
     concurrent-ruby (1.1.6)
     cookbook-omnifetch (0.9.1)
       mixlib-archive (>= 0.4, < 2.0)
-    cookstyle (5.23.0)
-      rubocop (= 0.75.1)
+    cookstyle (6.0.19)
```
tas50, line 355: 6.3 is out now

tyler-ball: I pulled in Cookstyle 6.3 and Chefstyle 1.0.2

tas50: Thanks

PR #1079, opened 2020-04-08
Explicitly build bcrypt_pbkdf-ruby

marcparadise:
On Windows, using the default behavior of installing the gem
only through the `gems` software definition installs an incompatible
version of the native extension for the gem (2.0-2.3 compat) while
directly building the gem ensures that the native extension is
compatible with the installed ruby version.

Fixes chef/chef-apply#129


omnibus/config/projects/chef-workstation.rb
```diff
@@ -85,6 +93,7 @@
   dependency "ruby-windows-devkit"
   dependency "ruby-windows-devkit-bash"
   dependency "ruby-windows-system-libraries"
+  # The default gem-install of
```
marcparadise, line 96: well remove

marcparadise:
Have confirmed that chef-run in this build is able to connect with an ed25519 key.

PR #1090, opened 2020-04-10
Patch docker-api to use CGI.escape to silence deprecation warning

### LINKED ISSUE #1040: When Workstation moves to Ruby 2.7 this april kitchen dokken will throw deprecations
Ruby 2.7 throws deprecations when using the URI.escape as it's been replaced. The docker-api gem which kitchen-dokken relies on uses this and throws lots of deprecation warnings under Ruby 2.7

```
-----> Starting Test Kitchen (v2.4.0)
/usr/local/lib/ruby/gems/2.7.0/gems/docker-api-1.34.2/lib/docker/container.rb:346: warning: URI.escape is obsolete
Chef container's Chef / Ohai release:
/usr/local/lib/ruby/gems/2.7.0/gems/docker-api-1.34.2/lib/docker/container.rb:346: warning: URI.escape is obsolete
Chef Infra Client: 16.0.131
/usr/local/lib/ruby/gems/2.7.0/gems/docker-api-1.34.2/lib/docker/container.rb:346: warning: URI.escape is obsolete
Ohai: 16.0.9
/usr/local/lib/ruby/gems/2.7.0/gems/docker-api-1.34.2/lib/docker/container.rb:346: warning: URI.escape is obsolete
rake, version 13.0.1
/usr/local/lib/ruby/gems/2.7.0/gems/docker-api-1.34.2/lib/docker/container.rb:346: warning: URI.escape is obsolete
Bundler version 2.1.2
/usr/local/lib/ruby/gems/2.7.0/gems/docker-api-1.34.2/lib/docker/container.rb:346: warning: URI.escape is obsolete
Successfully installed appbundler-0.13.2
Successfully installed appbundle-updater-1.0.3
2 gems installed
/usr/local/lib/ruby/gems/2.7.0/gems/docker-api-1.34.2/lib/docker/container.rb:346: warning: URI.escape is obsolete
```

clintoncwolfe:
Signed-off-by: Clinton Wolfe <clintoncwolfe@gmail.com>

## Description

Adds an omnibus config for the docker-api gem and provides a 2-line patch against the gem to silence a deprecation warning that appears under ruby 2.7

## Related Issue

Fixes #1040

## Types of changes

## Checklist:


tas50:
This feels like the wrong way to accomplish this. Traditionally we fork the repo (https://github.com/chef/docker-api) and then source our own gem in the Gemfile. That way we can maintain future fixes. This repo also needs to be updated for chef/chef test and inspec packages so doing it in Git allows us to reduce the amount of duplicate work.

tyler-ball:
@tas50 
> then source our own gem in the Gemfile

This would mean creating a `chef-docker-api` gem (or something similar) and uploading it to Rubygems, correct? Just want to ensure I understand what you are saying - that we would release and maintain a gem, rather than using an omnibus definition to build from a patch file or a fork.

> This repo also needs to be updated for chef/chef test and inspec packages

Can you help me understand better what this work is and when it would need to be done?

tas50:
If you just source this gem in the workstation gemfile then it becomes part of the package. No need to upload it to rubygems. No need to apply a brittle patch that will fail if any upstream changes are made. Just inject our own copy of the gem into the omnibus package. This is something we've done a lot in the past with chef-server upstream deps. This also allows to be good players in the OSS world because we can actually cut a PR to the upstream from that repo. By applying a patch we're hiding the fix in a way that it'll never actually get fixed upstream.

tyler-ball:
I thought using a Gemfile to source gems would break Appbundler? Or is that only for top level gems that Appbundler is... appbundling?

tas50:
That only breaks things that need a bin stub. Should be fine here

tyler-ball:
@clintoncwolfe I switched to using the Gemfile to point at our fork since I was wrong and this does not break Appbundler. I kicked off an ad-hoc build - would you confirm tomorrow that your changes got pulled in correctly and we do not end up with 2 copies of the `docker-api` gem?

@tas50 Bumping the dependencies pulled in a major version bump for `knife-vcenter` and `knife-cloud`. It looks like the only changes that warranted a major version bump here were to switch from Ruby >= 2.4 to >= 2.5. Since Chef Workstation currently uses 2.6 this seems fine to pull these changes in. Do you know of any issues with that? If so, I will pin them until we get ready to do the Chef Workstation major version release.

tas50:
The new knife-cloud loads the fog-core gem instead of fog gem which was a breaking change. The updates to the other projects were to use those new gems. knife-cloud is technically a breaking change, but it's a breaking change for things installed into Workstation which is a questionable use case at this point. From an end user standpoint nothing changed there. If you don't want to bring it in at this point you can just pin those back, but if you're doing the major release it should probably all go in.

PR #1139, opened 2020-04-29
Update the list of Kitchen Drivers

tas50:
Make sure we don't mention kitchen-docker anymore, add dokken and azure
and make sure we note which ones are in Workstation.


tas50:
@kagarmoe Where is that located?

tas50:
It should be fixed now @kagarmoe

PR #1143, opened 2020-05-01
test, ignore if it shows up at chef/chef-workstation

marcparadise:


marcparadise:
Thank you for testing.

PR #1164, opened 2020-05-09
Bring in Chef 16 and PJC that can build on Ruby 2.7

tyler-ball:
## Types of changes

- [x] Chore (non-breaking change that does not add functionality or fix an issue)

## Checklist:


components/gems/Gemfile
```diff
@@ -82,7 +82,7 @@ group(:omnibus_package) do
 
   # knife plugins
   gem "knife-acl", ">= 1.0.6"
```
tas50, line 84: knife-acl is built into chef 16 so this one can go away

components/gems/Gemfile
```diff
@@ -101,7 +101,7 @@ group(:omnibus_package) do
   gem "chef-sugar"
```
tas50, line 101: We should consider removing this since we ship this in chef 15.X more or less

components/gems/Gemfile
```diff
@@ -48,8 +48,8 @@ group(:omnibus_package) do
   # gems to Rubygems now, so letting this float on latest should always give us the latest
   # stable release. May have to re-pin around major version bumping time, or during patch
   # fixes.
-  gem "chef", "= 15.10.12 "
-  gem "chef-bin", "= 15.10.12 "
+  gem "chef", "= 16.0.287 "
+  gem "chef-bin", "= 16.0.287 "
   gem "ohai", ">= 15"
```
tas50, line 53: This should be > 16 just in case some other dep tries to pull it down

components/gems/Gemfile
```diff
@@ -48,8 +48,8 @@ group(:omnibus_package) do
...
   gem "cheffish", ">= 14.0.1"
```
tas50, line 54: Might as well make this 15+ now

components/gems/Gemfile.lock
```diff
@@ -321,7 +325,7 @@ GEM
       concurrent-ruby (~> 1.0)
       ffi-yajl (~> 2.2)
       http (~> 4)
-    chef-utils (15.10.12)
+    chef-utils (16.0.287)
     chef-vault (4.0.1)
```
tas50, line 329: There's a chef-vault dep in the gemfile that can go away now that it's a dep of chef itself

components/gems/Gemfile
btm: to clarify what @tas50, we've put functionality in chef-utils, which ships in core chef, that is mostly equivalent to chef sugar starting around 15.5+

PR #1165, opened 2020-05-11
chef-sugar helpers are now in chef-utils

btm:
chef-utils (from chef/chef/chef-utils) is included as part of core chef
now and includes most of the helpers that used to be separately shipped
in the chef-sugar library. We no longer ship the chef-sugar library and
thus no longer need to test it.


tas50:
I think we're good to pull this from workstation at this point. We should definitely think about why we shipped it in the DK / workstation releases originally. I can't think of why we'd need it there in the first place.

lamont-granquist:
I think I added it to DK originally and it was just because a large number of people used it

PR #1171, opened 2020-05-12
Take out the Ruby 2.4+ gate in the omnibus def

tas50:
There's no need for this anymore.


omnibus/verification/verify.rb
```diff
@@ -194,10 +194,10 @@ def components
 
       add_component "chef-apply" do |c|
         c.gem_base_dir = "chef-apply"
-      #   c.unit_test do
```
tyler-ball, line 197: Can you try uncommenting these and running an ad-hoc build? I found some of our other commented out verifications have started passing recently.

PR #1198, opened 2020-05-14
Pin chef-telemetry to avoid extra http deps

tas50:
This should also let us nuke the ext dirs again. This obviously requires some ad hocs and maybe we still have to include the ext dirs, but here hoping.


omnibus/Gemfile
```diff
@@ -5,7 +5,6 @@ gem "omnibus-software", git: "https://github.com/chef/omnibus-software.git", bra
 gem "artifactory"
 
 gem "pedump"
-gem "bcrypt_pbkdf", ">= 1.1.0.rc1"
```
marcparadise, line 8: We get this through `chef` anyway, right?

tas50: yep the deps we need for the build are all via chef in the gemfile.lock in omnibus.

PR #1209, opened 2020-05-17
Push a git tag after a VERSION change

btm:
Normally expeditor does a 'git tag' for us when we use
'built_in:bump_version' but in this case we are doing the version bump
on our own since sometimes we do a major/minor bump based on the the
date.

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -44,6 +44,9 @@ merge_actions:
         - "Expeditor: Skip Omnibus"
         - "Expeditor: Skip All"
       only_if: bash:.expeditor/determine_version.sh
+  - bash:.expeditor/push-git-tag.sh:
+      only_if: bash:.expeditor/determine_version.sh
+      post_commit: true
 
 pipelines:
   - verify:
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -30,6 +30,7 @@ github:
 # https://expeditor.chef.io/docs/getting-started/subscriptions/#merge-actions
 # The `merge_actions` key is a shortcut for the
 # `pull_request_merged:<GITHUB_REPO>:<RELEASE_BRANCH>:*` subscription.
+#
 merge_actions:
   - bash:.expeditor/determine_version.sh:
       ignore_labels:
@@ -39,14 +40,15 @@ merge_actions:
       ignore_labels:
         - "Expeditor: Skip Changelog"
         - "Expeditor: Skip All"
+  # The git commit happens here
+  - bash:.expeditor/push-git-tag.sh:
+      only_if: bash:.expeditor/determine_version.sh
+      post_commit: true
   - trigger_pipeline:omnibus/release:
       ignore_labels:
         - "Expeditor: Skip Omnibus"
         - "Expeditor: Skip All"
       only_if: bash:.expeditor/determine_version.sh
-  - bash:.expeditor/push-git-tag.sh:
-      only_if: bash:.expeditor/determine_version.sh
-      post_commit: true
 
 pipelines:
   - verify:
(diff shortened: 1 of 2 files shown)


.expeditor/config.yml
```diff
@@ -44,6 +44,9 @@ merge_actions:
         - "Expeditor: Skip Omnibus"
         - "Expeditor: Skip All"
       only_if: bash:.expeditor/determine_version.sh
+  - bash:.expeditor/push-git-tag.sh:
```
tduffield, line 47: So the tag is on the remote when the omnibus build occurs, we'll want to run this prior to triggering the omnibus build.

btm: I'm presuming the `post_commit: true` here is redundant?

.expeditor/config.yml
```diff
@@ -39,6 +39,8 @@ merge_actions:
       ignore_labels:
         - "Expeditor: Skip Changelog"
         - "Expeditor: Skip All"
+  - bash:.expeditor/push-git-tag.sh:
```
tyler-ball, line 42: The order in the config is the order the commands will be executed in, correct? I tried to confirm that in https://expeditor.chef.io/docs/getting-started/action-sets/ but I didn't see it.

If my understanding is correct, looks like BTM got this updated to run before the release.

However, the `bash` action is pre-commit by default (https://expeditor.chef.io/docs/reference/actions/#bash) so I think we need `post_commit: true` added back

btm: I think I actually had it here originally, but with the `post-commit: true` and expeditor had errored saying that the actions had to be listed in the order they would be executed, which had caused me to move it to the end.

It seems unlikely that the `trigger_pipeline:omnibus/release` action would be firing before a commit was finished, but it doesn't have a post-commit property on it.

I'm starting to get confused about what commit we're actually referring to.

tduffield: It should be before `trigger_pipeline`, and specified with a `post_commit: true`. I'll see if I can find out why it was giving you an error.

tduffield: I put your exact scenario into my spec tests and everything came up golden. Let's try your config with the following `merge_actions`:

```
            - bash:.expeditor/determine_version.sh:
                ignore_labels:
                  - "Expeditor: Skip Version Bump"
                  - "Expeditor: Skip All"
            - built_in:update_changelog:
                ignore_labels:
                  - "Expeditor: Skip Changelog"
                  - "Expeditor: Skip All"
            - bash:.expeditor/push-git-tag.sh:
                only_if: bash:.expeditor/determine_version.sh
                post_commit: true
            - trigger_pipeline:omnibus/release:
                ignore_labels:
                  - "Expeditor: Skip Omnibus"
                  - "Expeditor: Skip All"
                only_if: bash:.expeditor/determine_version.sh
```

tduffield: As for "what commit we're actually referring to", there is an implicit commit between the pre-commit and post-commit phases. That's likely what we're referring to.

btm: I probably had it between `bash:.expeditor/determine_version.sh:` and ` built_in:update_changelog:` before.

btm:
looks correct, thanks folks.

```
btm@btm-thinkpad:~/src/chef-workstation (master)$ git log -n 1
commit 42f02ad470b7e0b37728846fd8a9526c7a6e0bd7 (HEAD -> master, tag: 20.5.42, origin/master, origin/HEAD)
Author: Chef Expeditor <expeditor@chef.io>
Date:   Mon May 18 19:29:50 2020 +0000

    Executed '.expeditor/determine_version.sh'
    
    Obvious fix; these changes are the result of automation not creative thinking.
btm@btm-thinkpad:~/src/chef-workstation (master)$ cat VERSION 
20.5.42
```

tyler-ball:
Thanks @btm!

PR #1193, opened 2020-05-14
Fixing indentation issue on third-party-package pipeline

tyler-ball:
## Description
Error from buildkite:
```
2020-05-14 16:38:51 WARN   POST https://agent.buildkite.com/v3/jobs/b94f1321-bd21-465c-adf8-bee6fe85de54/pipelines: 422 `executor` is not a valid property on the `command` step. Please check the documentation to get a full list of supported properties. (Attempt 1/60 Retrying in 5s)
--
  | 2020-05-14 16:38:51 ERROR  Unrecoverable error, skipping retries
```

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Chore (non-breaking change that does not add functionality or fix an issue)

## Checklist:

- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have run the pre-merge tests locally and they pass.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] All commits have been signed-off for [the Developer Certificate of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

PR #1239, opened 2020-05-31
Temporarily comment out pipeline that publishes our Homebrew PR

tyler-ball:
We are actively troubleshooting how to do this in
https://github.com/chef/chef-workstation/pull/1193 but are running into
issues. We have the ChefConf release coming up in a few days so I am
going to comment this out and we will add it back after the ChefConf
release. We can manually create the Homebrew PR for this release.


tyler-ball:
@marcparadise I'm thinking we do this for now so that the ChefConf release goes _smoother_ and then we can figure out the promotion pipeline after Tuesday

PR #1266, opened 2020-06-11
Update InSpec, Train, ChefSpec, and others to the latest

tas50:
- Removes the pin on FFI < 1.13 since 1.13.1 resolves the issue we had
- chef-cli with updates the the chefignore file generated in `chef generate cookbook`
- ChefSpec with additional Chef 16 support and fixes for really annoying
Ruby 2.7+ warnings
- Train with the ffi 1.13 pin removed now that 1.13.1 is out plus fixes
  to support the new net-ssh which improves our ssh support in knife
ssh, knife bootstrap, inspec etc
- New knife tidy with additional options to control how many cookbooks to keep and a fix for a failure due to a missing template file
- New InSpec with some nice fixes


components/gems/Gemfile
```diff
@@ -17,9 +17,6 @@
 
 source "https://rubygems.org"
 
-# We found a segfault in ffi 1.13
-gem "ffi", "< 1.13.0"
```
tyler-ball, line 21: Can you also remove this pin from `omnibus/Gemfile`?

tas50: Will do.

tas50: Fixed

PR #1288, opened 2020-06-25
Add the habitat binary to the Workstation package

marcparadise:
We want to ensure that Workstation includes all of the tools required to
manage infrastructure - Chef Habitat is a key part of that.

- This adds a software definition to include the pre-built
  Chef Habitat 'hab' binary for each supported platform at WS build-time,
  in the same way we bundle workstation-app. We can visit switching this
  to build the binary instead of downloading pre-built in future
  iterations.
- It adds an update script to update the definition to the latest stable
  version of habitat.
- it adds an expeditor workload to trigger that script when a stable hab
is published.
- it adds a post-inst step that checks for conflicting `hab` install and
warns the operator thta WS should now manage the install.

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -71,6 +71,9 @@ subscriptions:
       # on this artifact's availability
       - trigger_pipeline:third-party-packages
 
+  - workload: habitat-sh/habitat:master_completed:project_promoted:habitat-sh/habitat:master:stable:*
+    actions:
+      - bash:.expeditor/update_habitat.sh
   - workload: chef/chef-workstation-app:master_completed:pull_request_merged:chef/chef-workstation-app:master:*
     actions:
       - bash:.expeditor/update_chef-workstation-app_to_latest.sh
--- a/.expeditor/update_habitat.sh
+++ b/.expeditor/update_habitat.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+#
+# Copyright:: Copyright (c) 2020 Chef Software Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+############################################################################
+# What is this script?
+#
+# Chef Workstation uses Expeditor (internal CI bot) to manage the bundled
+# version of the the Chef Habitat CLI. This script will be run by Expeditor
+# when a new version of habitat is published to the stable channel, to ensure that
+# Workstation is updated to and validated with the most recent stable version of
+# Chef Habitat.
+############################################################################
+
+set -evx
+
+BASE_URL="http://packages.chef.io/files/stable/habitat/latest"
+winsha=$(curl "$BASE_URL/hab-x86_64-windows.zip.sha256sum" --silent | cut -d ' ' -f 1)
+darwinsha=$(curl "$BASE_URL/hab-x86_64-darwin.zip.sha256sum" --silent | cut -d ' ' -f 1)
+linsha=$(curl "$BASE_URL/hab-x86_64-linux.tar.gz.sha256sum" --silent | cut -d ' ' -f 1)
+sw_def_file="omnibus/config/software/habitat.rb"
+
+MANIFEST_URL="$BASE_URL/manifest.json"
+MANIFEST=$(curl --silent "$MANIFEST_URL")
+version=$(jq -r '.version'<<< "$MANIFEST")
+
+sed -i -r "s/windows_sha= .*/windows_sha = \"$winsha\"/" $sw_def_file
+sed -i -r "s/linux_sha= .*/linux_sha = \"$linsha\"/" $sw_def_file
+sed -i -r "s/darwin_sha = .*/darwin_sha = \"$darwinsha\"/" $sw_def_file
+sed -i -r "s/^version \".*/version \"$version\"/" $sw_def_file
+
+branch="expeditor/hab-${version}"
+git checkout -b "$branch"
+git add $sw_def_file
+git commit --message "Bump habitat to $version." --message "This pull request was triggered automatically via Expeditor when habitat v. $version was merged to master." --message "This change falls under the obvious fix policy so no Developer Certificate of Origin (DCO) sign-off is required."
+open_pull_request
+
+# Get back to master and cleanup the leftovers - any changed files left over at the end of this script will get committed to master.
+git checkout -
+git branch -D "$branch"
+
--- a/omnibus/config/software/habitat.rb
+++ b/omnibus/config/software/habitat.rb
@@ -1,0 +1,57 @@
+#
+# Copyright:: Copyright (c) 2020 Chef Software Inc.
+# License:: Apache License, Version 2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+name "habitat"
+default_version "latest"
+license :project_license
+skip_transitive_dependency_licensing true
+
+# DO NOT MODIFY: version and checksums are populated by workstation/.expeditor/update_habitat.sh
+version "1.6.56"
+linux_sha = "a87f4ff7558f23724289e2c5a9b75920da364818167f63b26411be5a8b344800"
+darwin_sha = "42b6c417b88351e0b8ce99cb6ebcc104bbb5bc819ffc795f980d1201068035e6"
+windows_sha = "648157f1db680233b676725d3142ef118e5a77da000209a532fb54ff153b57dd"
+# END DO NOT MODIFY
+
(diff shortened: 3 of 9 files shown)


.expeditor/config.yml
```diff
@@ -80,6 +80,10 @@ subscriptions:
       # on this artifact's availability
       - trigger_pipeline:third-party-packages
 
+  # TODO is this a supported workload?
+  - workload: artifact_published:stable:habitat:{{version_constraint}}
```
tduffield, line 84: `artifact_published` is only for Artifactory artifacts. You can see how you can subscribe to Habitat promotions here: https://github.com/chef/automate/blob/master/.expeditor/config.yml#L172-L176

omnibus/config/software/habitat.rb
```diff
@@ -0,0 +1,38 @@
+name "habitat"
+default_version "master"
+license "Apache-2.0"
```
tduffield, line 3: The source code is apache licensed, but the binary is not. The binary uses Chef's MLSA.

omnibus/config/software/habitat.rb
```diff
@@ -0,0 +1,38 @@
+name "habitat"
+default_version "master"
```
tduffield, line 2: I think you'd want to use `latest` here vs `master`. Once again, you're installing a pre-compiled binary and not installing from source.

omnibus/config/software/habitat.rb
marcparadise: It looks like "Chef EULA" is the right one from the description: https://docs.chef.io/chef_license/#chef-eula

omnibus/package-scripts/chef-workstation/postinst
```diff
@@ -25,7 +25,21 @@ else
...
 
-binaries="chef-run berks chef chef-cli chef-analyze chef-apply chef-shell chef-solo chef-vault cookstyle delivery foodcritic inspec kitchen knife ohai push-apply pushy-client pushy-service-manager chef-client"
+binaries="chef-run berks chef chef-cli chef-analyze chef-apply chef-shell chef-solo chef-vault cookstyle delivery foodcritic inspec kitchen knife ohai push-apply pushy-client pushy-service-manager chef-client hab"
+
+old_hab_path=$(which hab)
+
+if [[ "$old_hab_path" != "" ]]; then
+    loc=$(readlink "$old_hab_path") # If it's not a link readlink gives back the same path.
+    if [[ $loc != *"chef-workstation"* ]] ; then
+      echo ""
```
marcparadise, line 35: Phrasing check request for @chef/docs-team. 

Context:  This is expected to be a minimum occurrence/edge case.   This message will be shown after installing the updated package on mac/linux, if the operator has already installed Habitat  directly on the node.  When that happens, we prefer for customers to receive future Hab updates through the Workstation updates, in the same way we deliver Infra, Inspec, etc as part of Workstation. 



If a node has both Workstation and Habitat installed,  it simplifies life for the operator and for us if future updates to habitat are provided via Workstation updates.  Nothing will break if they go back to updating habitat directly from stable, current, etc - but each Workstation install will overwrite it.  
If habitat

.expeditor/update_habitat.sh
```diff
@@ -0,0 +1,56 @@
...
+
+sed -i -r "s/windows_sha= .*/windows_sha = \"$winsha\"/" $sw_def_file
+sed -i -r "s/linux_sha= .*/linux_sha = \"$linsha\"/" $sw_def_file
+sed -i -r "s/darwin_sha = .*/darwin_sha = \"$darwinsha\"/" $sw_def_file
+sed -i -r "s/^version \".*/version \"$version\"/" $sw_def_file
+
+branch="expeditor/hab-${version}"
+git checkout -b "$branch"
+git add $sw_def_file
+git commit --message "Bump habitat to $version." --message "This pull request was triggered automatically via Expeditor when Delivery CLI $version was merged to master." --message "This change falls under the obvious fix policy so no Developer Certificate of Origin (DCO) sign-off is required."
```
clintoncwolfe, line 50: Delivery CLI?

marcparadise: Hah, nice catch.

omnibus/package-scripts/chef-workstation/postinst
```diff
@@ -25,7 +43,25 @@ else
...
+        ;;
+      *)
+        echo ""
+        echo "NOTE: Chef Habitat on this node is now managed by Chef Workstation."
+        echo "      Future Workstation updates will always include the latest stable"
+        echo "      Habitat release."
+        echo ""
+        ;;
+    esac
+    rm -rf "$old_hab_path"
```
tyler-ball, line 63: I'm conflicted on this one. It would surprise me if I installed Visual Studio Code and that decided to uninstall notepad++ for me. I realize that we own both Chef Workstation and Habitat but it still seems weird...

omnibus/package-scripts/chef-workstation/postinst
```diff
@@ -25,7 +43,25 @@ else
     PREFIX="/usr"
 fi
 
-binaries="chef-run berks chef chef-cli chef-analyze chef-apply chef-shell chef-solo chef-vault cookstyle delivery foodcritic inspec kitchen knife ohai push-apply pushy-client pushy-service-manager chef-client"
+binaries="chef-run berks chef chef-cli chef-analyze chef-apply chef-shell chef-solo chef-vault cookstyle delivery foodcritic inspec kitchen knife ohai push-apply pushy-client pushy-service-manager chef-client hab"
+
+old_hab_path=$(which hab)
+
+if [ "$old_hab_path" != "" ]; then
+    loc=$(readlink "$old_hab_path") # If it's not a link readlink gives back the same path.
```
marcparadise, line 51: Will change this to usr PREFIX/bin/hab  instead of looking to detect it anywhere in the path.

omnibus/package-scripts/chef-workstation/postinst
```diff
@@ -25,7 +43,25 @@ else
     PREFIX="/usr"
 fi
 
-binaries="chef-run berks chef chef-cli chef-analyze chef-apply chef-shell chef-solo chef-vault cookstyle delivery foodcritic inspec kitchen knife ohai push-apply pushy-client pushy-service-manager chef-client"
+binaries="chef-run berks chef chef-cli chef-analyze chef-apply chef-shell chef-solo chef-vault cookstyle delivery foodcritic inspec kitchen knife ohai push-apply pushy-client pushy-service-manager chef-client hab"
+
+old_hab_path=$(which hab)
+
+if [ "$old_hab_path" != "" ]; then
```
marcparadise, line 50: `if [ -f "$PREFIX/bin/hab" ] ; then`

marcparadise: Suggested: if symlink replace; if executable rename (hab.old)

omnibus/package-scripts/chef-workstation/postinst
tyler-ball: Updating for a note made during manual review - I am in favor of overwriting an existing symlink in `$PREFIX/bin` and renaming an existing binary from `$PREFIX/bin/hab` to `$PREFIX/bin/hab.old`

.expeditor/update_habitat.sh
```diff
@@ -0,0 +1,56 @@
...
+# Chef Workstation uses Expeditor (internal CI bot) to manage the bundled
+# version of the the Chef Habitat CLI. This script will be run by Expeditor
+# when a new version of habitat is published to the stable channel, to ensure that
+# Workstation is updated to and validated with the most recent stable version of
+# Chef Habitat.
+############################################################################
+
+set -evx
+
+BASE_URL="http://packages.chef.io/files/stable/habitat/latest"
```
tyler-ball, line 32: I'm only used to requesting info from `omnitruck.chef.io` - who maintains `packages.chef.io`?

tyler-ball, line 32:
```suggestion
BASE_URL="https://packages.chef.io/files/stable/habitat/latest"
```

marcparadise:
This has been updated with the discussed changes around replacing hab safely in systems where it's already installed. Will run it through the pipeline and test the post-install behaviors again tomorrow.

omnibus/package-scripts/chef-workstation/postinst
```diff
@@ -25,7 +43,39 @@ else
...
+  loc=$(readlink "$hab_target_path")
+  case $loc in
+    *chef-workstation*)
+      # We own this path, so we're not taking it over from another install.
+      ;;
+    *)
+      takeover="1"
+  esac
+elif [ -f "$hab_target_path" ]; then
+  takeover="1"
```
tyler-ball, line 59:
```suggestion
```

Don't we only want to rename if the file exists?

omnibus/package-scripts/chef-workstation/postinst
```diff
@@ -25,7 +43,39 @@ else
...
+  echo "NOTE: Chef Habitat on this node is now managed by Chef Workstation."
+  echo "      Future Workstation updates will always include the latest stable"
+  echo "      Habitat release."
+  echo ""
+fi
+
+if [ "$rename" = "1" ]; then
+  mv "$hab_target_path" "$hab_target_path.orig"
+  echo "'$hab_target_path' has been renamed to '$hab_target_path.orig'"
+  echo ""
```
tyler-ball, line 75:
```suggestion
  echo ""
  echo "NOTE: Chef Habitat on this node is now managed by Chef Workstation."
  echo "      Existing install $hab_target_path' has been renamed to"
  echo "      '$hab_target_path.orig'"
  echo ""
```

omnibus/package-scripts/chef-workstation/postinst
```diff
@@ -25,7 +43,39 @@ else
...
+    *)
+      takeover="1"
+  esac
+elif [ -f "$hab_target_path" ]; then
+  takeover="1"
+  rename="1"
+fi
+
+if [ "$takeover" = "1" ]; then
+  rm -rf "$hab_target_path"
```
tyler-ball, line 64:
```suggestion
```

What do you think about, in the symlink case, not doing anything because it will be overwritten by the `ln -sf` below?

omnibus/package-scripts/chef-workstation/postinst
marcparadise: For context, the 'now managed' message will be displayed already;  this is an additional info message that wil be shown after that. 

I've changed the wording of the rename message in the commit coming up though, `Original habitat binary has been moved to '$hab_target_path.orig'"

PR #1301, opened 2020-06-29
Enable fast installs on Windows

tas50:
We use this in Chef itself.

1 	No system restore point is saved for this installation.
2 	Perform only File Costing and skip checking other costs.
4 	Reduce the frequency of progress messages.


tas50:
I tested this on windows 10 and it was noticeably faster. I need to test it on older Windows releases to ensure they still install.

tas50:
Confirmed working on 2012 RTM which has the oldest release of Windows installer

PR #1303, opened 2020-06-30
Remove the omnibus install of rb-fsevent

tas50:
This should fix double gems being installed on Mac systems


tas50:
This totally didn't work since the new gem release on april 30th still ships the old binary.

PR #1362, opened 2020-07-22
Create ~/.chef when `chef` command invoked. 

### LINKED ISSUE #810: Workstation should create .chef dirs in the user home by default
After installing workstation every user needs a .chef dir, but we don't create it for them. If there isn't a .chef dir in the home dir then we should create it.

### LINKED ISSUE #1355: Update chef command to create ~/.chef directory
## Description

Follow up to https://github.com/chef/chef-workstation/pull/924

*nix systems install packages as `root` so there is no good way to have them create a directory in `$HOME` for the user installing the package. Because of this, we want to add the logic for ensuring `~/.chef` exists into the main-chef-wrapper. This will ensure it exists any time a user executes `chef`.

## Acceptance criteria

1. `chef` command updated to check for the existence of `~/.chef` on any invocation
1. If it does not exist, it gets created
1. If it does exist, do nothing
1. This logic works for `chef -v` and `chef -h` and no-arg-no-flag `chef` invocations

## Notes

We're starting to add multiple pre-flight checks to the `chef` command (license acceptance checking, chef-apply telemetry check, certain knife plugins, etc.). This is starting to add a lot of startup time and we duplicate these checks in a lot of different places. We need to refactor these different concerns into a single pattern for adding fast pre-flight checks to our software.

marcparadise:
This will create the ~/.chef dir cross-platform whenever that directory does not exist.  Many of our tools expect that directory, and will dire-seeming errors/warnings when we invoke `chef` and it's not present.   

fixes #1355


components/main-chef-wrapper/main.go
```diff
@@ -106,6 +113,15 @@ Available Commands:
 	fmt.Printf(msg)
 }
 
+func doStartupTasks() error {
+	dir, err := homedir.Expand("~/.chef")
+	if err != nil {
+		return err
+	}
+	err = os.MkdirAll(dir, 0700)
```
tyler-ball, line 121: In some weird case where they don't have a home directory and we create it for them, do you think there are any issues with `700` permissions on that directory? I would think that would be desired but my default on Darwin is `755` (which I am sure is because of my umask).

Just want to make sure we are intentional about this setting.

marcparadise: I'll play it safe and skip creation if we can't find an existing valid home directory.

marcparadise: Final version - because it's safe for us to continue running even if we don't make the directory, I updated this to just try to make the dir (using Mkdir, not MkdirAll) and move on.

Issue #1354, opened 2020-07-21
Add hab version into chef -v


tyler-ball:
Also update verify.rb scripts to check that it can be ran correctly
Also update the tray app to show it in the about page

PR #1363, opened 2020-07-22
Update default_version of habitat during build

marcparadise:
This will ensure that the omnibus build manifest reflects the
correct version number instead of 'latest'.

### DIFF
--- a/.expeditor/update_habitat.sh
+++ b/.expeditor/update_habitat.sh
@@ -42,7 +42,7 @@ version=$(jq -r '.version'<<< "$MANIFEST")
 sed -i -r "s/windows_sha= .*/windows_sha = \"$winsha\"/" $sw_def_file
 sed -i -r "s/linux_sha= .*/linux_sha = \"$linsha\"/" $sw_def_file
 sed -i -r "s/darwin_sha = .*/darwin_sha = \"$darwinsha\"/" $sw_def_file
-sed -i -r "s/^version \".*/version \"$version\"/" $sw_def_file
+sed -i -r "s/^default_version \".*/default_version \"$version\"/" $sw_def_file
 
 branch="expeditor/hab-${version}"
 git checkout -b "$branch"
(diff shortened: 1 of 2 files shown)


marcparadise:
Part of  chef/chef-workstation#1354

marcparadise:
> It is... weird that our `version-manifest.json` contains the default version instead of the version we specify but ¯_(ツ)_/¯
> 
> How about we get rid of `version` and only specify `default_version`? Then just have expeditor update that since we are the only consumers of this omnibus definition
Should work.  Running the verify pipeline with that change.

tyler-ball:
Also for posterity: here is an example of the `version-manifest.json` output that is causing us to make this change:
```
"docker-api": {
  "locked_version": "865a72d16204f731780630240d69df88832dfadb",
  "locked_source": {
    "git": "https://github.com/chef/docker-api.git"
  },
  "source_type": "git",
  "described_version": "master",
  "license": "MIT"
},
"habitat": {
  "locked_version": "latest",
  "locked_source": {
    "url": "https://packages.chef.io/files/stable/habitat/latest/hab-x86_64-darwin.zip",
    "sha256": "42b6c417b88351e0b8ce99cb6ebcc104bbb5bc819ffc795f980d1201068035e6"
  },
  "source_type": "url",
  "described_version": "latest",
  "license": "project_license"
},
```

`docker-api` turns its described_version into a SHA because omnibus knows how to convert a git tag to a SHA.

But I don't know why, even though we were overriding `default_version` with `version`, we only see the `default_version` in the manifest but ¯\_(ツ)_/¯

marcparadise:
This latest change is building/verifying. I'll check back on it later and will merge if no issues.

tyler-ball:
@marcparadise did you see [my comment](https://github.com/chef/chef-workstation/pull/1363#pullrequestreview-457121056) about changing the `source url`?

marcparadise:
Yes, it's in there now and verified.

PR #1368, opened 2020-07-23
Fix ordering of purge built-in

marcparadise:
Corrects validation error because default ordering of purge is after commit
but default ordering of bash is pre-commit.

ref: https://expeditor.chef.io/docs/reference/status-checks/#config-002


marcparadise:
Fix already in master

PR #1388, opened 2020-07-31
Pull in smaller omnibus sofware defs + expand gem cleanup

tas50:
Add some new cleanup we're doing in chef/chef that shaves a few more files from the install
Bring in the new omnibus-software
  - libxml2 without debug support
  - libxslt without debug and profiler support


tas50:
Adhoc of this running here: https://buildkite.com/chef/chef-chef-workstation-master-omnibus-adhoc/builds/328#d7d0ee77-de9b-41d1-a01c-e20f8cc43137

PR #1396, opened 2020-08-04
Docs overhaul

IanMadd:
## Description

This is a bit of docs overhaul.

- Migrates ctl_chef.md from chef-web-docs.
- Removes all shortcodes from chef-workstation.
- Updates README and makefile
- Remove some templates that aren't necessary or used

**Note**: Netlify is not going to be able to build this PR preview because there are changes that I have to make with the theme in chef-web-docs after this PR and another PR in `desktop-config` are merged.

## Related Issue

We're removing the shortcodes chef-workstation (and other repos) because they're not really working as intended.

The intent of shortcodes is that we have reusable text that will be identical no matter where it is added. The problem is that we are documenting the same things across multiple repos. We came up with the practice of adding shortcodes that are used in repos other than `chef-web-docs` by adding a prefix to each shortcode file name depending on the repository that they're added to, but we've ended up with multiple shortcodes that have slightly different names but the same text. 

This creates problems if the text in a shortcode is updated in one repository, but the matching text is not updated in matching shortcodes in other repositories. The other issue we have is that some shortcodes are only used once in their repositories. Both of these issues defeats the purpose of shortcodes, which is that we have reusable text that can be used in multiple places and is identical no matter where it is added.

So now we'll just keep all shortcodes in `chef-web-docs/themes/docs-new/layouts/shortcodes` and all changes to that text should be made there.

## Types of changes

## Checklist:

- [X] All commits have been signed-off for [the Developer Certificate of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).


tyler-ball:
```
Error: Error building site: "/workdir/www/content/workstation/_index.md:17:1": failed to extract shortcode: template for shortcode "chef_workstation" not found
--
  | Makefile:37: recipe for target 'lint' failed

```

From the Buildkite failure. Do we need to re-fetch the git submodule? Or is this work still in flight on the `chef-web-docs` side?

IanMadd:
@tyler-ball There are changes I have to make in chef-web-docs after this PR is merged.

marcparadise:
@IanMadd is there any way to get the deploy preview passing?

IanMadd:
@marcparadise That should take care of it.

PR #1400, opened 2020-08-06
Build azure compute and network SDKs from our fork

tas50:
This should reduce about 2.5 million lines of SDK code we install.


marcparadise:
@tas50 is this still relevant? I know you had some related  work going on down other paths.

tas50:
If Microsoft doesn't do their thing I think we still want to consider this to speed up the overall workstation install time and make the upgrades faster on Windows. If we expect people to upgrade every 2 weeks we can't take 20 minutes to upgrade.

PR #1413, opened 2020-08-11
Fix Homebrew PR creation

marcparadise:
* switch to the macos executor, because the non-ES linuxbrew image is
  not a permitted to access expeditor secrets.
* Updates buildkite emoji alias to be :mac:
* mac-compatibility fixes for sed usage.
* Fix PR request to set maintainer_can_modify to false. This otherwise
  fails the PR with a message about the non-existent field
  `fork_collab`, possibly because the account creating the branch
  (chef-ci) doesn't have proper push access to the branch, triggering
  an odd validation failure from GH API.  Ref: https://github.com/octokit/rest.js/issues/490
* fix indentation issue in third party package pipeline yml (tball,
squashed)

### DIFF
--- a/.expeditor/create_pr_for_homebrew.sh
+++ b/.expeditor/create_pr_for_homebrew.sh
@@ -1,31 +1,33 @@
 #!/bin/bash
 
-# This bumps the version in the ENV var 'EXPEDITOR_VERSION' on the project specified
-# It fetches the sha256 via omnitruck api
-
 set -ex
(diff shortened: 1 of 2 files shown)


marcparadise:
Testing result , shows the created PR: https://github.com/marcparadise/homebrew-cask-autotesting/pull/2 (picture too, I will prboably be deleting that repo): 



capture_112

PR #1435, opened 2020-08-20
Update all deps to the latest with performance optimizations

tas50:


tas50:
Doing #1445 instead

PR #1439, opened 2020-08-24
Bring Chef Rollout Client To Worksation

mudash:
Signed-off-by: Mudassar Shafique <15931574+mudash@users.noreply.github.com>

## Description

This PR brings roll-out client for policy changes from automate repository to chef-workstation. Adds some basic framework for testing. Adds it into Buildkite and Omnibus pipelines for Chef Workstation.

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly.


components/chef-automate-collect/commands/metadata_collectors.go
```diff
@@ -0,0 +1,240 @@
...
+	PolicyDomainURL  string `json:"policy_domain_url"`
+	Description      string `json:"description"`
+	SCMMetadata
+	CiMetadata
+}
+
+type SCMMetadata struct {
+	policyLockPath         string
+	SCMType                string `json:"scm_type"`
+	SCMWebType             string `json:"scm_web_type"`
```
marcparadise, line 35: I think SCM-related fields are changing to not be explicitly "SCM",  but it might be we need to re-sync after Dan has completed the related WIP.

mudash:
> I think we'll need to do work on some future refactoring to bring this more in line with our other CLI tools (and perhaps some ideas can be taken back from here into our existing tools ), but the shape of it looks right and the build-related components look good.

Exactly, I think the same!

components/chef-automate-collect/commands/metadata_collectors.go
tyler-ball: Correct, we'll need to re-sync this after those changes. There is some future work we need to do around getting integration tests setup so we can see when the API change causes failures.

components/chef-automate-collect/VERSION
```diff
@@ -0,0 +1 @@
+0.0.1
```
tyler-ball, line 1:
```suggestion
0.1.0
```

components/chef-automate-collect/chef-automate-collect.go
```diff
@@ -0,0 +1,7 @@
+package main
```
tyler-ball, line 1: The Golang convention is to use `_` in the filename instead of `-`. Will you update this and the other files that have dashes in them?

components/chef-automate-collect/go.mod.bak
```diff
@@ -0,0 +1,10 @@
+module github.com/chef/chef-workstation/components/chef-automate-collect
```
tyler-ball, line 1: Lets leave this file out. We don't need to revert back to this for any reason and I don't see anything online about a `go.mod.bak`.

omnibus/package-scripts/chef-workstation/postinst
```diff
@@ -77,7 +77,7 @@ fi
 
 
 
-binaries="chef-run berks chef chef-cli chef-analyze chef-apply chef-shell chef-solo chef-vault cookstyle delivery foodcritic inspec kitchen knife ohai push-apply pushy-client pushy-service-manager chef-client hab"
+binaries="chef-run berks chef chef-cli chef-automate-collect chef-analyze chef-apply chef-shell chef-solo chef-vault cookstyle delivery foodcritic inspec kitchen knife ohai push-apply pushy-client pushy-service-manager chef-client hab"
```
tyler-ball, line 80:
```suggestion
binaries="chef-run berks chef chef-cli chef-analyze chef-apply chef-shell chef-solo chef-vault cookstyle delivery foodcritic inspec kitchen knife ohai push-apply pushy-client pushy-service-manager chef-client hab"
```

I propse that we don't link these onto the path because only `chef push` is meant to call that and it should be able to figure out the `embedded/bin` folder

omnibus/package-scripts/chef-workstation/postrm
```diff
@@ -13,7 +13,7 @@ cleanup_symlinks() {
   # Keep removed symlinks in this list, so that removal of upgraded packages still cleans up
   # leftovers from older versions.
   workstation_binaries="berks chef chef-cli chef-apply chef-shell chef-solo chef-vault cookstyle dco delivery foodcritic inspec kitchen knife ohai push-apply pushy-client pushy-service-manager chef-client"
-  binaries="chef-run chef-workstation-app $workstation_binaries chef-analyze hab"
+  binaries="chef-run chef-workstation-app $workstation_binaries chef-analyze chef-automate-collect hab"
```
tyler-ball, line 16:
```suggestion
  binaries="chef-run chef-workstation-app $workstation_binaries chef-analyze hab"
```

components/chef-automate-collect/VERSION
mudash: Done

components/chef-automate-collect/chef-automate-collect.go
mudash: Done

components/chef-automate-collect/go.mod.bak
mudash: Deleted `go.mod.bak`

omnibus/package-scripts/chef-workstation/postinst
mudash: Latest tech design discussions in the team suggest we will not modify chef push, but instead chef cli wrapper will be modified to make a call to chef push and then to chef-automate-collect. We can discuss if that is not the case.

tyler-ball: Even if we are calling `chef-automate-collect` from the `chef` wrapper, we can still reference it from `/opt/chef-workstation/embedded/bin` instead of `/usr/local/bin` (which is where these symlinks end up).

I don't want to hold up this PR anymore but I would like to reconsider these symlinks when we wire up the `chef` wrapper to call `chef-automate-collect`

PR #1465, opened 2020-09-04
WIP - mirror command structure from chef-cli

marcparadise:
h2. Description 

This PR  sets up `main-chef-wrapper` to shadow every top-level command in `chef-cli`, including flags and help text. This will let us swap out individual commands that we don't want to forward to `chef-cli` by editing the command definition, and without needing to worry about modifying a routing table to determine where to send each command. 

This uses `cobra` to provide the argument/flag processing. `cobra` is the preferred library for managing CLI command flags/arguments/help under golang in Workstation and Automate. This replaces the simplified switch/case mechanism we were using previously, because we're about to outgrow that with upcoming work for Rollouts, and moving the `shell-init` command out of ruby to better support Windows.  

As a side effect, inclusion of parameters and help text in the shadowed commands means that we will be able to provide command help from `main-chef-wrapper` in most cases, resulting in a signficant speedup when someone only wants to view help. Another benefit is that `shell-init`'s autocomplete setup is something we get for "free" out of cobra, with support for bash/zsh/fish/powershell. 

h2. Further notes

What I don't like is that we have to duplicate some the command structure/setup already defined in `chef-analyze` so that autocomplete will work correctly for `report` and `.  Some of this (like the infra_flags that will become consistent/common for all commands that needs them) should get pushed down into chef/go-libs or similar.  More specfic things - like the command structure and flags for report/capture - could get exposed from chef-analyze.  That would also would make the implicit dependency from main-chef-wrapper -> chef-analyze explicit. 

Some trouble areas still in flight: 

- unknown flags don't pass through, which is a problem for commands like `chef gem` 
- a lot of inconsistent wording found when transcribing command descriptions. I fixed some, but we need to 
take a close look. Particular areas of concern:  
    - policyfile/Policyfile/Policyfile.rb/POLICYFILE/POLICY_FILE 
    - ^ all of the above, with '.lock'. 
- Some flags are getting made consistent which will result in changes. '-d' is reserved for '--debug'.  '-c' is reserved for '--config'.  This results in a couple of minor changes for consistency noted in the code. 


components/main-chef-wrapper/cmd/capture.go
```diff
@@ -0,0 +1,57 @@
...
+`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return passThroughCommand(dist.AnalyzeExec, "", os.Args[1:])
+		},
+	}
+)
+
+func init() {
+	// TODO - this is a change for compatibilty with global "d"ebug flag.
+	//        this code is "D", we will need to update main-chef-wrapper with this change.
```
marcparadise, line 47: s/main-chef-wrapper/chef-analyze/

marcparadise:
Testing notes: to verify, I've been running each command and comparing output from main-chef-wrapper and . This may be good to add as an integration test, so that we can catch discrepancies until we can automate  keeping them in sync.

nikhil2611:
Blockages handled here: https://github.com/chef/chef-workstation/pull/2130
Unit Test cases: https://github.com/chef/chef-workstation/pull/2157

vkarve-chef:
All relevant changes merged through https://github.com/chef/chef-workstation/pull/2289

PR #1479, opened 2020-09-17
Update Chef Infra Client to 16.5

tas50:
Also bump the minimums for other libs to speed up depsolving


tyler-ball (approved):

tas50:
This breaks tests

PR #1482, opened 2020-09-28
WS powershell on windows does not require runas

### LINKED ISSUE #1449: Remove Powershell Admin requirement
Chef Workstation tools do not typically require administrative access on Windows.  

Exceptions:
- `chef-client`, if someone is running it out of WOrkstation to converge their local node - not recommend practice; but if so, they can still `run as administrator` to launch a shell that will provide the necessary permissions. 
 
- `hab` - attempting to start supervisor requires escalated permissions. However, if it's run in a standard shell without escalated permissions, UAC prompts for the necessary permissions and the action completes successfully if they are granted. 

To remove admin access requirments, replace 'runas' with 'open': 

https://github.com/chef/chef-workstation/blob/master/omnibus/resources/chef-workstation/msi/assets/start-chefws.ps1#L12-L16

To be aware of: this will cause the powershell prompt to appear to take longer. This may be because the `chef shell-init` is continuing to run in the background while windows UAC , but more research should be done here to see if we can speed this up. On a modern machine, the delay between the PowerShell window opening and text appearing/input available was ~ 10 seconds. 

Reference: https://github.com/chef/chef-workstation/issues/1365

marcparadise:
This changes the verb used to launch powershell and cmd on
Windows from 'runas' to 'open'.  This reduces the permissions required,
so that operators are no longer shown the UAC/admin prompt
that displays when starting WS powershell on Windows.

Note that the UAC prompt on Windows concealed a slow startup
time for the WS powershell, because it uses `chef shell-init` to
configure the environment.  This will be addressed separately.

Fixes #1449 


tas50:
Paging @mwrock

mwrock (approved):
I think we used to have the `runas` verb because we had so many customers run into various incarnations of `Access Denied` errors that opening an admin shell put them on a better path for success. However, even without it, one can easily "opt in" to admin mode by right clicking the shortcut and selecting Run as Administrator so this seems sensible for giving non admins a way to use the shell.

tyler-ball (approved):

PR #1524, opened 2020-11-03
Remove Foodcritic and Stove binaries

tas50:
Remove these EOL tools now that we have Cookstyle for code linting and knife supermarket share, which now respects the chefignore file.

This change brings in a new chef_deprecations gem, which provides simple binaries for foodcritic and stove to exit 1 with deprecation messages. That way we can give users the nudge to the new tools.


lamont-granquist (approved):

lamont-granquist (approved):

robbkidd (approved):
goodbye old tools

PR #1556, opened 2020-11-23
Remove macOS 10.13 from the build matrix

tas50:
We support N-2 for macOS releases


marcparadise (approved):

marcparadise:
Let's bring up to date before merge

PR #1559, opened 2020-11-25
roll out reporting

mudash:
Signed-off-by: Mudassar Shafique <15931574+mudash@users.noreply.github.com>

## Description

This includes set of changes that:
- add the chef-automate-collect call after the Chef Policy call in the main wrapper if the environment variables are set appropriately
- retries the http request in case of internal server error
- refactors a little bit for chef cli wrapper flow

## Types of changes

## Checklist:


components/main-chef-wrapper/main.go
```diff
@@ -47,6 +49,21 @@ func main() {
 	case "report", "capture":
 		cmd = exec.Command(dist.AnalyzeExec, allArgs...)
 
+	case "push", "push-archive":
+		cmd = exec.Command(dist.WorkstationExec, allArgs...)
```
marcparadise, line 53: Let's update this so that if Rollouts are enabled but not configured correctly in validateRolloutSetup, we fail with the error message before we perform the push.   When `CHEF_AC_ROLLOUT_ENABLED` is set, the operator intent of using Rollouts is clear - so if it's not possible to use them, it should be an error.

components/main-chef-wrapper/main.go
```diff
@@ -47,6 +49,21 @@ func main() {
...
+		runCmd(cmd)
+		if validateRolloutSetup() {
+			policyGroup := fmt.Sprintf("-g%s", allArgs[1])
+			policyLockFile := fmt.Sprintf("-l%s", allArgs[2])
```
marcparadise, line 57: formatted strings should probably be quoted in output, in case of something like a filename with spaces.

marcparadise:
In `report_new_rollout.go`, the messaging for error codes 500 and 501 is incorrect - could you update that to let those pass through to the default error handler in the case?

marcparadise (changes requested):

components/main-chef-wrapper/main.go
mudash: That sounds good. I have addressed it so we will error out and will not do the policy push if the required setup isn't complete.

components/main-chef-wrapper/main.go
mudash: Used a different approach in the updates now.

mudash:
> In `report_new_rollout.go`, the messaging for error codes 500 and 501 is incorrect - could you update that to let those pass through to the default error handler in the case?

This mechanism was sitting there from original version, I have changed it to route to the default case.

components/main-chef-wrapper/main.go
```diff
@@ -43,14 +46,24 @@ func main() {
...
+	switch getAction(subCommand) {
 	case "report", "capture":
 		cmd = exec.Command(dist.AnalyzeExec, allArgs...)
 
+	case "policy-rollout":
+		cmd = exec.Command(dist.WorkstationExec, allArgs...)
+		runCmd(cmd)
+		allArgs = []string{"report-new-rollout", "-g", allArgs[1], "-l", allArgs[2],
+			"-s", os.Getenv("CHEF_AC_SERVER_URL"), "-u", os.Getenv("CHEF_AC_SERVER_USER")}
+		cmd = exec.Command(dist.AutomateCollectExec, allArgs...)
```
marcparadise, line 58: For future iterations, I wonder if we can do this more transactionally - first, create the rollout record successfully in automate, then update the results after we execute the command.  This way even if the final update fails, we're guaranteed to have _some_ record of the rollout before we proceed to push the policy.

components/main-chef-wrapper/main_test.go
```diff
@@ -0,0 +1,69 @@
+//
+// Copyright 2019 Chef Software, Inc.
```
marcparadise, line 2: ^ 2021

components/main-chef-wrapper/main_test.go
```diff
@@ -0,0 +1,69 @@
...
+	os.Setenv("CHEF_AC_SERVER_USER", "testuser")
+	os.Setenv("CHEF_AC_AUTOMATE_URL", "http://testhost2")
+	os.Setenv("CHEF_AC_AUTOMATE_TOKEN", "xyz123455677709u0")
+	got := validateRolloutSetup()
+	assert.Equal(t, got, true)
+}
+
+func Test_validateRolloutSetup_Invalid(t *testing.T) {
+
+	os.Setenv("CHEF_AC_SERVER_URL", "http://testhost")
```
marcparadise, line 38: Ideally we'd test for each var being recognized as required, and not just the one.

marcparadise (changes requested):
Thanks @mudash . This looks like a good place from which to build. I've left  a couple of minor comments inline.

components/main-chef-wrapper/main_test.go
mudash: Sure, more is better, updated.

components/main-chef-wrapper/main_test.go
mudash: Updated. Good reminder :)

marcparadise (approved):

PR #1580, opened 2020-12-09
Enable git caching for omnibus builds

tas50:
Speed up our build times


marcparadise (approved):
This looks good, though I don't think it will help in our CI environment because our builders start clean every time.

PR #1616, opened 2020-12-24
Pin rake to avoid double rake, bump git, bump Test Kitchen

tas50:
We're getting double rake deps that results in gem warnings.


tas50:
https://buildkite.com/chef/chef-chef-workstation-master-omnibus-adhoc/builds/384#40b41ee3-0554-46d9-ac8e-6f35e6505917

PR #1617, opened 2020-12-24
Don't define libarchive twice in the omnibus config

tas50:
There's no need to do this.


tas50:
https://buildkite.com/chef/chef-chef-workstation-master-omnibus-adhoc/builds/385#fd372e63-2cfd-4ba5-aa74-eb8dccf1db68

PR #1761, opened 2021-02-20
Fix Chef Client update PRs lacking a name

tas50:
It's a bit of a hack, but it should work (right?)


.expeditor/config.yml
```diff
@@ -86,6 +86,8 @@ subscriptions:
       - bash:.expeditor/update_chef-analyze_to_latest.sh
   - workload: artifact_published:stable:chef:16*
     actions:
+      # hack to make sure update_dep creates a proper PR title
+      - export EXPEDITOR_GEM_NAME="Chef Infra Client"
```
tduffield, line 90: So, this won't work. 

Could we maybe switch to subscribing to the gem publish? It would functionally be the same.

tduffield (changes requested):

.expeditor/config.yml
tas50: I think the reason we didn't do the gem publish is that we want to subscribe to just chef (not chef-*) and not chef 15. In the current config we pin to the chef 16 promotes.

tduffield:
@tas50 I think https://github.com/chef/chef-workstation/pull/1777 will work a bit better.

PR #1819, opened 2021-03-11
Publish macOS builds for M1 macs

tas50:
Publish native packages for M1 macs


tas50:
2nds of this is on master now. No need to keep this around

PR #1876, opened 2021-04-05
Update openSSL on macOS to 1.1.1k to mac infra client

tas50:
We're going to need this when we produce an M1 package, but it also
enables TLS 1.3 for things like curl which is a good idea overall.


tas50:
https://buildkite.com/chef/chef-chef-workstation-master-omnibus-adhoc/builds/397#c8b75781-e43c-4301-be30-85a4dc25d3b9

tas50:
Fails due to delivery not compiling on OpenSSL 1.1.1

PR #2010, opened 2021-05-24
WIP upgrade workstation to ruby 3.0 (DO NOT MERGE)

lamont-granquist:


components/gems/Gemfile.lock
```diff
@@ -715,12 +722,13 @@ GEM
       net-ssh (>= 2.6.5)
       net-ssh-gateway (>= 1.2.0)
     netaddr (1.5.1)
-    nokogiri (1.10.10)
-      mini_portile2 (~> 2.4.0)
-    nokogiri (1.10.10-x64-mingw32)
-      mini_portile2 (~> 2.4.0)
-    nokogiri (1.10.10-x86-mingw32)
-      mini_portile2 (~> 2.4.0)
+    nokogiri (1.11.0.rc4)
```
tas50, line 725: How did we end up with an old rc here.

components/gems/Gemfile
```diff
@@ -21,7 +21,7 @@ group(:omnibus_package, :development, :test) do
   # Require a minimum version that is packaged in the Ruby install, so we do
   # not install old unecessary versions. When we bump Ruby we need to look
   # at these pins and adjust them.
-  gem "minitest", "= 5.13.0"
+  gem "minitest", "= 5.14.2"
 
   gem "rake", "= 13.0.1" # prevent double rake as of 12.2020
```
tas50, line 26: I think we can remove this entirely now that we've cleaned up some of our omnibus mess

PR #2022, opened 2021-05-25
Remove the unused rb-fsevent-gem patch

tas50:
We're not using this anywhere


tas50:
https://buildkite.com/chef/chef-chef-workstation-master-omnibus-adhoc/builds/448#0f7360dd-21ea-4fc7-b48b-6c73d11bf7c0

PR #2041, opened 2021-06-02
knife configure actually creates a credentials file

tas50:
We updated this to match the new behavior


kagarmoe (approved):
tenor-26743051

PR #2042, opened 2021-06-02
Remove the mention of the chef-vault cookbook and gem installation

tas50:
The cookbook is dead now that we ship the resources / helpers in Chef
Infra Client itself. Also why are we talking about installation in the
Workstation docs. That's just a waste of space.


kagarmoe (approved):
tenor-122545720

PR #2059, opened 2021-06-10
Collapse the requirements down to a single section

tas50:
We ship a single package and the app isn't optional anymore. This made
no sense in its own section.


kagarmoe (approved):
tenor-161674365

PR #2060, opened 2021-06-10
Use more reasonable ram/disk requirements

tas50:
You would not have a good time at all with 2GB


kagarmoe (approved):
tenor-67374538

PR #2069, opened 2021-06-11
Rollback unf_ext to 0.0.7.6

lamont-granquist:
The 0.0.7.7 version isn't compatible with ruby-3.0 on windows (which
makes it a bit of a puzzle that this even built and shipped...)

(I also sorta doubt this "works" on windows, but there's nothing we can do about it, and
this causes the gem depsolver to not immediately puke)

### DIFF
--- a/components/gems/Gemfile.lock
+++ b/components/gems/Gemfile.lock
@@ -990,9 +990,7 @@ GEM
     uber (0.1.0)
     unf (0.1.4)
       unf_ext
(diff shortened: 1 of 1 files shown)


tas50:
can we pin this with a note so we don't roll it forward again in a week?

lamont-granquist:
it shouldn't roll forwards i had to hack it up manually

PR #2130, opened 2021-07-06
Invalid flags and command handled

nikhil2611:
## Description
Handled the invalid command and flags blockages.


components/main-chef-wrapper/cmd/root.go
```diff
@@ -93,11 +94,11 @@ func init() {
 	// These flags are common to all child commands.  Some of them do not need config or debug,
 	// so we can look at pushing this down; but it seems to make sense since it's present for more
 	// commands than it isn't.
-	rootCmd.PersistentFlags().StringVarP(&options.configFile, "config", "c", "", "Read configuration from `CONFIG_FILE_PATH`")
-	rootCmd.PersistentFlags().StringVar(&options.licenseAcceptance, "chef-license", "",
-		"Accept product license, where `ACCEPTANCE` is one of 'accept', 'accept-no-persist', or 'accept-silent'")
-	rootCmd.PersistentFlags().BoolVarP(&options.debug, "debug", "d", false,
-		"Enable debug output when available")
+	// rootCmd.PersistentFlags().StringVarP(&options.configFile, "config", "c", "", "Read configuration from `CONFIG_FILE_PATH`")
```
marcparadise, line 97: I am not clear what the changes are between commented/uncommented versions; but in any case we should remove the commented-out code.

components/main-chef-wrapper/cmd/capture.go
```diff
@@ -36,6 +36,7 @@ var (
 Captures a node's state as a local chef-repo, which can then be used to
 converge locally.
 `,
+		DisableFlagParsing: true,
```
marcparadise, line 39: Good find with disableFlagParsing; but I would expect that we specify this only on commands that accept arguments that we can't reasonably shadow in our wrappers. It seems like we're using it in more places than that?

components/main-chef-wrapper/cmd/root.go
```diff
@@ -61,7 +61,7 @@ var (
 		// Stop framework from showing default errors. This prevents duplicate errors or
 		// unncessary info from showing on passthrough commands; and
 		// allows us control over error rendering for internal commands.
-		SilenceErrors: true,
```
marcparadise, line 64: This comment is now out of date.  In addition, was there a specific type of error that we were suppressing that has become a problem? 

 The intent was to make sure the actual error is not obscured in front of a wall of usage text/instructions - it should be clearly shown as the final piece of output from a failed command whenever possible. 

If silencing errors caused other issues, we should look for an approach that allows them to be handled, but still doesn't show additional information after the error message - that's the most important thing for the user to see.

components/main-chef-wrapper/cmd/diff.go
```diff
@@ -32,7 +32,7 @@ var (
 	pager     bool
 
 	diffCmd = &cobra.Command{
-		Use:                   "diff [POLICYFILE] [--head | --git GIT_REF | POLICY_GROUP | POLICY_GROUP...POLICY_GROUP]",
+		Use:                   "diff [Policyfile] [--head | --git GIT_REF | POLICY_GROUP | POLICY_GROUP...POLICY_GROUP]",
```
marcparadise, line 35: I think there's probably some more change to Policyfile/policyfile/POLICYFILE usage.  Perhaps a general guide could be: 

- `Policyfile` - referring to a specific file named Policyfile
- `POLICYFILE` - referring to a user-supplied argument that names a policyfile. If referring to the argument in descriptive text, it should also use POLICYFILE.  This keeps it consistent with other argument placeholders across our commands.  You can see this in this example with GIT_REF, POLICY_GROUP, etc. 
- `policyfile` - referring to the non-specific policyfile as in "a policyfile" or "open the policyfile".  

/cc @tas50 @chef/docs-team

marcparadise (changes requested):

components/main-chef-wrapper/cmd/root.go
nikhil2611: I removed this because these flags are global here but  these flags are not applicable to many commands in chef-cli, but anyhow adding these flags again here as this is a bug in chef-cli and these are the high level flags should be applicable to all the commands in chef-cli. I raised an issue for the same and will pick after this.

components/main-chef-wrapper/cmd/root.go
nikhil2611: The 2 issues I am facing if we are silencing the error is
1. When we are passing the invalid command at that time It fails in the  execute function  on   `rootCmd.Execute()`  call, without calling the chef-cli repo and exits the OS without any error.
2. Also in some cases where the commands don't accept the arguments so we are setting `DisableFlagParsing` to false. Now if I am passing any invalid flag to this command,  it will not parse this as an argument and checks for the passed invalid flag in the file and when it dosen't find it just exits the OS without any error when the execute func gets called..

components/main-chef-wrapper/cmd/diff.go
nikhil2611: Thanks. Updating this.

components/main-chef-wrapper/cmd/capture.go
nikhil2611: Yes I agree with this, but if I am not specifying this to the commands like `clean-policy-cookbooks` which don't accept the arguments than when I am passing any invalid flags to these command  these flags are not getting parsed as an argument and the cobra tries to find the invalid flag here itself without letting it to invocate chef-cli and exits the OS on the Execute function without any error.

nikhil2611:
Changes merged in main with PR --> #2333

PR #2243, opened 2021-08-27
Try appbundling rspec as a "chefspec" command

tas50:
This way people can find it.


tas50:
This doesn't work

Issue #2153, opened 2021-07-15
Investigate moving `chef env` to the go CLI


tas50:
As a user I want `chef` commands to run quickly
so that I can get the information I need to do my job quickly

In order to make the Chef CLI as performant as possible we should investigate moving the `chef env` command from the chef-cli Ruby app to the chef Go app. This could potentially greatly speed up this command on Windows systems.

PR #2430, opened 2021-12-07
Move chef env to go

i5pranay93:
## Description
we move the chef env command from the chef-cli Ruby app to the chef Go app. This speeds up this command on Windows systems.

## Related Issue
solves https://github.com/chef/chef-workstation/issues/2153

## Types of changes

## Checklist:

### DIFF
--- a/components/main-chef-wrapper/cmd/env.go
+++ b/components/main-chef-wrapper/cmd/env.go
@@ -23,6 +23,7 @@ import (
 	"os"
 
 	"github.com/chef/chef-workstation/components/main-chef-wrapper/dist"
+	"github.com/chef/chef-workstation/components/main-chef-wrapper/platform-lib"
 	"github.com/spf13/cobra"
 )
 
@@ -31,14 +32,19 @@ var envCmd = &cobra.Command{
 	Short:              "Prints environment variables used by %s",
 	DisableFlagParsing: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		ReturnEnvironment()
+		if len(args) >= 1 {
+			cmd.Help()
+			os.Exit(0)
+		}else {
+			ReturnEnvironment()
+		}
 		return nil
 	},
 }
 
 
 func ReturnEnvironment() {
-	err := platform_lib.Environment()
+	err := platform_lib.RunEnvironment()
 	if err != nil {
 		fmt.Fprintln(os.Stderr, "ERROR:", err.Error())
 		os.Exit(4)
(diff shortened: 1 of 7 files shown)


components/main-chef-wrapper/main.go
```diff
@@ -42,6 +47,34 @@ func createDotChef() {
...
+//	}
+//	todo ==> incase home directory is needed we can add it to string
+	installationPath :=  platform_lib.ExpectedOmnibusRoot() + "/ruby-env.json"
+	fmt.Printf(installationPath)
+	result, err := exists(installationPath)
+	if err != nil {
+		log.Fatalf(err.Error())
+	}
+	if result == true {
+		fmt.Print("file exists======== ruby script not needed")
```
marcparadise, line 64: I think we have to take this a step further.  The contents of this file can change across version, so we'll need to regenerate it when a new version of WS is in use. The easiest way I can think of to address that is to capture the workstation version in the JSON file, and compare it if the file is present.

components/main-chef-wrapper/main.go
```diff
@@ -42,6 +47,34 @@ func createDotChef() {
...
 }
 
+func createRubyEnv(){
+//	get installation path
+//	homepath, err := homedir.Dir()
+//	if err != nil {
+//		log.Fatalf(err.Error())
+//	}
+//	todo ==> incase home directory is needed we can add it to string
+	installationPath :=  platform_lib.ExpectedOmnibusRoot() + "/ruby-env.json"
```
marcparadise, line 57: I think we have to resolve and use `~/.chef-workstation/ruby-env.json`. The operator usually won't be running `chef` with root permissions, so creating this file would fail.

omnibus/config/software/ruby-env-manifest.rb
```diff
@@ -0,0 +1,29 @@
...
+
+name "ruby-env-manifest"
+
+skip_transitive_dependency_licensing true
+license :project_license
+
+source path: "#{project.files_path}/#{name}"
+
+build do
+  ruby "#{project_dir}/default/chef_cli_paths_as_json.rb #{install_dir}/ruby-env.json"
```
marcparadise, line 28: THis might be tricky.  We want to bundle exec this in the context of the installed environment to get the values; and IIRC omnibus has some bundler-busting behavior that might interfere? 

Too, is there a chance of getting different results on the actual installed version, vs the builder? 

What if we invoke this using our embedded bundler/ruby during `postinst`, and the windows equivalent?  It's  a good fit, because this is global state that makes sense to capture once at install time.

components/main-chef-wrapper/main.go
i5pranay93: addressing in commit https://github.com/chef/chef-workstation/pull/2430/commits/5ed7f94e99ce333eb65021b553d5b327cc33aae8

components/main-chef-wrapper/main.go
i5pranay93: resolved https://github.com/chef/chef-workstation/pull/2430/commits/01e46e7fc070e67e3e62b8d01c9e1b3edb1c353b

PR #2434, opened 2021-12-09
Fix chef workstation app software definition

jeremiahsnapp:
Whitelist node-related binaries in the chef-workstation-app software definition

Fix copy's source path in chef-workstation-app software definition

### DIFF
--- a/omnibus/config/software/chef-workstation-app.rb
+++ b/omnibus/config/software/chef-workstation-app.rb
@@ -36,6 +36,17 @@ platform_name = if macos?
 source_url = "https://packages.chef.io/files/unstable/chef-workstation-app/#{version}/chef-workstation-app-#{version}-#{platform_name}.zip"
 app_install_path = "#{install_dir}/components/chef-workstation-app"
 
+# These electron dependencies are pulled in/created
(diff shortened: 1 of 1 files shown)


omnibus/config/software/chef-workstation-app.rb
```diff
@@ -36,6 +36,17 @@
 source_url = "https://packages.chef.io/files/unstable/chef-workstation-app/#{version}/chef-workstation-app-#{version}-#{platform_name}.zip"
 app_install_path = "#{install_dir}/components/chef-workstation-app"
 
+# These electron dependencies are pulled in/created
+# by this build. They may have dependencies that aren't met
+# on the install target - in which case the tray application
+# will not be runnable.  That does not affect the rest of
+# the chef-workstation installation, so we will whitelist the
+# dependencies to allow it to continue in any case.
+if linux?
```
tas50, line 45: Do we know what deps we're relying on and if we can ship those in Workstation?

tas50: Nevermind. I see we had these whitelisted before.

PR #2570, opened 2022-02-04
Move chef env to go (cherrypick)

i5pranay93:
## Description
we move the chef env command from the chef-cli Ruby app to the chef Go app. This speeds up this command on Windows systems.

## Related Issue
solves https://github.com/chef/chef-workstation/issues/2153
related https://github.com/chef/chef-workstation/pull/2430
## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly.


components/main-chef-wrapper/platform-lib/cli_helpers.go
```diff
@@ -0,0 +1,86 @@
...
+	//return os.Getenv("HOME")
+	home, err := os.UserHomeDir()
+	if err != nil {
+		log.Fatal(err)
+	}
+	return filepath.Join(home, dist.WorkstationDir)
+}
+
+func OmnibusGemRoot() string {
+	gemRoot := rubyenvMap["omnibus path"].(map[string]interface{})["GEM_ROOT"]
```
sanghinitin, line 48: either define struct for ruby env or check before going next can we convert that interface into particular type 
data, ok := rubyenvMap["omnibus path"].(map[string]interface{})
if ok {
gemRoot := data["GEM_ROOT"]

}
like this in all func

components/main-chef-wrapper/platform-lib/environment_darwin.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+func RunEnvironment() error {
+	// call all the environment info here.
+	envObj := WorkstationEnvInfo()
+	ymldump, err := yaml.Marshal(envObj)
+	if err != nil {
+		log.Fatalf("error: %v", err)
+	}
+	fmt.Printf("----:\n%s\n\n", string(ymldump))
```
sanghinitin, line 56: if you are using log then remove fmt

components/main-chef-wrapper/platform-lib/version_darwin.go
```diff
@@ -34,7 +35,7 @@ func init() {
 	manifestMap = manifestHash()
 }
 func Version() error {
-	if omnibusInstall() == true {
+	if OmnibusInstall() == true {
```
sanghinitin, line 38: you don't have to check true

components/main-chef-wrapper/platform-lib/version_darwin.go
```diff
@@ -140,5 +141,45 @@ func ExpectedOmnibusRoot() string {
...
 }
+
+func UnmarshallRubyEnv() map[string]interface{} {
+	filepath := path.Join(omnibusRoot(), "ruby-env.json")
+	jsonFile, err := os.Open(filepath)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "ERROR:", err.Error())
+		os.Exit(4)
+	}
+	byteValue, _ := ioutil.ReadAll(jsonFile)
```
sanghinitin, line 154: move defer here

components/main-chef-wrapper/platform-lib/environment_darwin.go
```diff
@@ -0,0 +1,162 @@
...
+
+	//fmt.Println(envObj)
+
+	// make yml dump like ruby --   ui.msg YAML.dump(info)
+	return nil
+}
+
```
sanghinitin, line 63: write some comments

components/main-chef-wrapper/platform-lib/environment_darwin.go
i5pranay93: log was giving timestamp(e.g--` 2022/02/08 11:11:35`), so fmt was more suitable.

components/main-chef-wrapper/lib/version_constants.go
```diff
@@ -0,0 +1,3 @@
+package lib
+
+const ChefCliVersion = "5.4.2"
```
tas50, line 3: Where is this used? Also this version isn't right

omnibus/files/ruby-env-script/ruby-env-script.rb
```diff
@@ -0,0 +1,98 @@
...
+info["omnibus root"] = omnibus_root
+info["ruby info"] = ruby_info
+info["build_version"] = chef_ws_build_version
+info["chef-cli"] = chef_cli_version
+
+j = JSON.pretty_generate(info)
+
+puts "Running ruby script to write environment path in ~/.chef-workstation"
+environment_file = ARGV[0]
+File.open(environment_file, "w") do |f|
```
tas50, line 96: You can probably simplify this whole block to just `File.write(whatever)`

components/main-chef-wrapper/lib/version_constants.go
i5pranay93: @tas50 I wonder why it still shows here, it was removed in later commit.

i5pranay93: Removing this, it was temporary, i was planning to bump i, like in cli.

components/main-chef-wrapper/cmd/env.go
```diff
@@ -31,10 +32,26 @@ var envCmd = &cobra.Command{
 	Short:              "Prints environment variables used by %s",
 	DisableFlagParsing: true,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return Runner.PassThroughCommand(dist.WorkstationExec, "", os.Args[1:])
+		if len(args) >= 1 {
+			cmd.Help()
+			os.Exit(0)
```
sanghinitin, line 37: is this else required here we can simply remove it.

components/main-chef-wrapper/main.go
```diff
@@ -42,6 +54,118 @@ func createDotChef() {
...
+			}
+		}
+	}
+
+	platform_lib.InitializeRubyMap()
+}
+
+func createRubyEnvWindows() {
+	InstallerDir := `C:\opscode\chef-workstation`
+	home, err := os.UserHomeDir()
```
sanghinitin, line 94: i think this func should return error when something goes wrong we are logging error but continuing work .

components/main-chef-wrapper/platform-lib/cli_helpers.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	return rubyExe
+}
+
+func RubyVersion() string {
+	rubyVersion := ""
+	if rubyenvMap == nil {
+		return rubyVersion
+	}
+	data, ok := rubyenvMap["ruby info"].(map[string]interface{})
```
sanghinitin, line 71: we are parsing everything dynamic if we already know what data will come we should use struct.it will save us checking and dynamic casting

components/main-chef-wrapper/platform-lib/version_darwin.go
```diff
@@ -136,9 +137,49 @@ func ExpectedOmnibusRoot() string {
...
+	defer jsonFile.Close()
+	var rubyenvHash map[string]interface{}
+	json.Unmarshal([]byte(byteValue), &rubyenvHash)
+	return rubyenvHash
+}
+
+func MatchVersions() bool {
+	// check version from env.json file and workstation version
+	home, err := os.UserHomeDir()
```
sanghinitin, line 162: error not handled

components/main-chef-wrapper/platform-lib/version_windows.go
```diff
@@ -147,3 +148,45 @@ func ExpectedOmnibusRoot() string {
 	rootPath = strings.Replace(s, "//", "/", -1)
 	return rootPath
 }
+
+func UnmarshallRubyEnv() map[string]interface{} {
```
sanghinitin, line 152: i think we should keep one func to unmarshal data it's not platform specific.read file then send byte to unmarshal.

sanghinitin (changes requested):

components/main-chef-wrapper/main.go
i5pranay93: Yeah, since we are running script with a check before every command, we just log the error, so that other command flow is not broken. and when env runs, the flow breaks. Even if something goes wrong in script which creates .json file another command should not be impacted.

components/main-chef-wrapper/platform-lib/cli_helpers.go
i5pranay93: since JSON data is unstructured, for now so i have used dynamic parsing.

components/main-chef-wrapper/platform-lib/version_windows.go
i5pranay93: Yes, we can do that. I had done that initially, but later when i realised path sperator were different for windows and unix i shifted it back.

i5pranay93: so i moved it to each platform so that each platform uses it's own unmarshall, based on path it has.

components/main-chef-wrapper/platform-lib/cli_helpers.go
sanghinitin: ok,got it.

components/main-chef-wrapper/platform-lib/version_windows.go
sanghinitin: ok, but we can pass file path and then we can use, but it's ok

sanghinitin (approved):

vkarve-chef (approved):

PR #3319, opened 2025-04-17
reverting this change due to chef oss limitations with secrets

sean-simmons-progress:
reverting: https://github.com/chef/chef-workstation/pull/3311 

build error:

```
/usr/local/bundle/gems/vault-0.16.0/lib/vault/client.rb:303:in `rescue in request': The Vault server at `https://vault.ps.chef.co' is not currently (Vault::HTTPConnectionError)
```

will need a workaround if we need this public.

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -18,7 +18,6 @@ changelog:
 
 pipelines:
   - verify:
-      public: true
       description: Pull Request validation tests
   - docker/build
   - omnibus/release:


nikhil2611 (approved):

sean-simmons-progress:
@jaymzh  FYI we are having to revert this due to build error limitation on access to secrets. Will need a workaround but for now reverting this so that we can at least still get tests ran.

jaymzh:
@sean-simmons-progress - can you assign an internal ticket to whoever should be working on fixing this? workstation is opensource, its verify pipeline should be public

PR #3549, opened 2026-04-30
Updated test-kitchen and all other components to the latest version

ashiqueps:
## Types of changes

## Checklist:

- [ ] If `Gemfile.lock` has changed, I have used `--conservative` to do it and included the full output in the Description above.


components/gems/Gemfile
```diff
@@ -60,37 +72,41 @@ group(:omnibus_package) do
...
   # fixes.
 
-  gem "chef", ">= 18.2"
-  gem "chef-bin", ">= 18.2"
-  gem "chef-utils", ">= 18.2"
+  if RUBY_PLATFORM.include?("mingw")
+    gem "chef", ">= 18.10.17", "< 19.0.0"
+  else
+    gem "chef", ">= 18.10.17", "< 19.0.0", platforms: [:ruby]
+  end
```
Stromweld: Why not just make it this?
```suggestion
  gem "chef", ">= 18.10.17", "< 19.0.0"
```

components/gems/Gemfile.lock
```diff
@@ -803,32 +793,36 @@ GEM
...
+      test-kitchen (>= 1.16, < 5)
       vsphere-automation-sdk (~> 0.4)
     kitchen-vra (3.3.3)
       ffi-yajl (>= 2.2.3, < 3.1.0)
       highline
       rack (>= 1.6, < 4.0)
       test-kitchen
       vmware-vra (~> 3.0, >= 3.2.0)
-    knife (18.10.18)
+    knife (18.6.13)
```
Stromweld: we may want to look at why this is being downgraded. That will mean it's missing latest fixes in knife.

components/gems/Gemfile.lock
```diff
@@ -1317,16 +1304,16 @@ GEM
...
       ffi (>= 1.16.0, < 1.18)
-      json (>= 1.8, < 3.0)
+      json (>= 2.19.2, < 3.0)
       mixlib-shellout (>= 2.0, < 4.0)
       net-scp (>= 1.2, < 5.0)
       net-ssh (>= 2.9, < 8.0)
     train-habitat (0.2.22)
-    train-kubernetes (0.2.1)
-      k8s-ruby (~> 0.16.0)
+    train-kubernetes (0.1.7)
```
Stromweld: not sure if this is an issue, but looks like this is also get downgraded.

components/gems/Gemfile
```diff
@@ -25,16 +25,28 @@ group(:omnibus_package, :development, :test) do
...
   gem "rdoc", "~> 6.4.1" # 6.4.1.1 required for CVE-2024-27281, allow patch upgrades
   gem "net-imap", ">= 0.2.5" # 0.2.5+ required for CVE fix - GHSA-j3g3-5qv5-52mj
+
+  # Ruby 3.1.x ships with stringio 3.0.1.2 as a default (built-in) gem. If bundler
+  # installs a newer version (e.g. 3.2.0 via psych), both versions are present and
+  # Gem::Specification.reset raises "Unresolved or ambiguous specs" for stringio,
+  # which causes `gem which <any-gem>` to fail on Windows (breaks the chef-dlls step
+  # that runs `gem which chef-powershell`). Pin to the exact version Ruby 3.1 ships with
+  # to prevent a second version from being installed alongside the default gem.
+  gem "stringio", "= 3.0.1.2"
```
Stromweld, line 49: My understanding is that is what appbundler is supposed to handle as it locks to specific versions and fixes the gem spec issue if multiple versions exist. In the gemfile.lock we are seeing this get downgraded from previous versions. I'm not sure if that is also what is causing other gems like knife or train-kubernetes to be downgraded as well.

components/gems/Gemfile.lock
ashiqueps: Fixed. While updating chef-zero got updated to latest and knife had a pin for chef-zero to the older version which downgraded knife.

components/gems/Gemfile
ashiqueps: Fixed.

components/gems/Gemfile.lock
ashiqueps: There is a dependency with k8s-ruby, when that got upgraded, due to the pin train-kubernates got downgraded. Fixed it by using the current version of k8s-ruby.

components/gems/Gemfile
ashiqueps: Yes, Appbundler will resolve these issues, but the build was failing on Windows before Appbundler. That's why we had to ping the version in the Gemfile to avoid the conflict. Also, Knife and train-kubernetes have been fixed separately.

nikhil2611 (approved):

